PR #3, opened 2014-11-06
Set server to listen on :: so both ipv4 and ipv6 connections can be made

prurigro:
By default, **node.js** only listens on **ipv4**, but it can be configured to listen on both **ipv4** _and_ **ipv6** if you specify `::` as the address.


cjdelisle:
Closing this because it seems not to be an issue anymore.

Issue #28, opened 2016-08-21
Typo in README?


ktorn:
I'm confused by this sentence in the README (under Security):

> It is acceptable for anyone who does not have the key to be able to change anything in the pad or add anything, even the server.

Isn't the whole pad encrypted, making it _impossible_ for anyone without the key to make changes?

cjdelisle:
basically we're saying that even though a person can't read the data, they can indeed change it.

ktorn:
OK, but any change would essencially destroy the pad (i.e. the clients cannot decrypt the change). Is this correct?

cjdelisle:
Oh actually this issue might be considered bunk because the hashes in ChainPad will fail to validate so trying to change the content will make the hash fail to validate

PR #29, opened 2016-08-22
clarify how mitm attack is mitigated

ktorn:
Added some extra clarification with regards to pad content changes by 3rd-parties without the private key. See #28


ansuz:
thanks!

kpcyrd:
Since this is about mitm, this part of the original readme is spot on: `or a script which plays Rick Astleys's greatest hits.`

If you're able to do that, you're able to get the key you need to make legit changes (and read the pad). You still need https (prevent backdoor insertion by the network), a honest server (prevent backdoor insertion by the server) and an xss free implementation (prevent backdoor insertion through bugs).

PR #39, opened 2016-09-27
Using express.static instead of custom handler

ochronus:
The original custom file handler code caused the server not sending the content-type header in some cases. This killed the site e.g. when using CloudFlare - safari tried to download the index.html instead of rendering it and chrome displayed the source code.

I might be missing something but I replaced that code with 2 simple express.static handlers - which handle content type (among other things) correctly

### DIFF
--- a/server.js
+++ b/server.js
@@ -13,12 +13,12 @@ var config = require('./config');
 config.websocketPort = config.websocketPort || config.httpPort;
 
 // support multiple storage back ends
-var Storage = require(config.storage || './storage/mongo');
+var Storage = require(config.storage||'./storage/mongo');
 
 var app = Express();
 app.use(Express.static(__dirname + '/www'));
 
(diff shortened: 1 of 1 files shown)


ansuz:
At a glance, I'm not sure why the tests are failing.

I pulled your branch and tested locally, and everything seems good, so maybe it's just a CI issue.

ochronus:
Yup it's a CI issue, the Chrome WebDriver is not installed there

ansuz:
I hate to nitpick when you're improving things, but it looks like quite a few lines were changed for stylistic purposes alone.

Would you mind restricting it to the functional modifications?

ochronus:
Please do "nitpick" :) I'm the new guy here and I'd like to accommodate to your coding style. Those changes were my editor being too smart automagically, sorry :) fix incoming

ansuz:
Excellent, thanks!

PR #63, opened 2016-12-05
How to run cryptpad as systemd daemon

pojebunny:
Code part should remain as proposed.
Merithroical part can be freely changed.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -39,7 +39,32 @@ cp config.js.dist config.js
 ## for example aon the default mongodb port `mongodb://localhost:27017/demo_database`
 $EDITOR config.js
 
+```
+
+## Running
+
+Run server.js using node:
 node ./server.js
+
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -39,7 +39,32 @@ cp config.js.dist config.js
...
+```
+[Unit]
+Description=cryptpad
+After=network.target
+
+[Service]
+ExecStart=/usr/bin/node /where/is/your/cryptpad/server.js
+Restart=always
+User=root (user that has access to all cryptpad related files, root recommended ONLY for testing)
+Group=root (group that has access to all cryptpad related files, root recommended ONLY for testing)
```
ansuz: Since root is only recommended for testing, there should probably be steps listed for how to create a new user. Ideally CryptPad installation and configuration should be as simple as possible.

ansuz:
closing since there has been no progress

PR #67, opened 2016-12-16
Translation to Brazilian Portuguese

gustavohmsilva:
Added header with contributor name and contact for future inquires regarding pt-br translation.
Legal stuff, bottom and header translated, links comments translated whenever possible.
Please inform me if new translating keys get's added, so I can maintain this translation updated.

Translation done with love from Brazil.


ansuz:
Thanks a lot! I just have one question.

I've never really looked into how the AGPL-3.0 and CC-4.0 interact, is it correct that to comply we only have to keep your header in the translation file?

ansuz:
In any case, it's merged into `soon`, which will be merged to master and deployed today.

PR #70, opened 2016-12-19
Reviewed and fine tuned FP and TOS (Brazilian Portuguese)

gustavohmsilva:
I've fixed some minor errors and revised some texts in order to provide better sense for the text. 
Beside that, I've reduced the text (keeping it's meaning of course) of the buttons close to the bottom of the front page, since on Firefox, Chrome and Safari, the buttons seems to not be adaptive oriented as the main text divs.


ansuz:
I'll deploy later today, thanks again :+1:

PR #72, opened 2016-12-29
adding docker and docker-compose

felixboehm:
Have built a docker image for cryptpad and want to contribute this back.

- [x] CLA signed
- [x] added some docs

Let me know if anything else needed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+data
+Dockerfile
+docker-compose.yml
+.git
+.gitignore
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,43 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && apt-get install -y \
+  vim \
+  wget \
+  git \
+  curl \
+  npm \
+  nodejs-legacy
+
+ARG VERSION=0.3.0
+
+# Download stable version
+# RUN wget https://github.com/xwiki-labs/cryptpad/archive /${VERSION}.tar.gz -O /cryptpad.tar.gz \
+#   && mkdir -p /cryptpad \
+#   && tar -xzf /cryptpad.tar.gz -C /cryptpad --strip-components=1 \
+#   && rm /cryptpad.tar.gz
+
+# Download from github
+# RUN git clone https://github.com/xwiki-labs/cryptpad.git
+
+# Add code directly
+ADD . /cryptpad
+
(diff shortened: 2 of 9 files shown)


felixboehm:
@tboerger please review

cryptpad-docker.md
```diff
@@ -0,0 +1,69 @@
...
+- Adding config.js to customize folder
+- Persistance for datastore and customize folder
+
+## TODO
+
+```
+cryptpad_1  | Linking config.js
+cryptpad_1  | Using secure websockets: true
+cryptpad_1  | Using storage adapter: './storage/file'
+cryptpad_1  | sed: -e expression #1, char 27: unknown option to `s'
```
felixboehm: The reason for this error is unclear to me. @tboerger ?

tboerger: Sed expects a backup suffix for the -i option. Write something like -i.bak and drop this .bak file afterwards

ansuz:
Looks good to me, but I know very little about Docker. I know of a few people who have done some CryptPad/Docker work independently, so I'll see about getting their input and hopefully we can converge on a single method for managing things.

ansuz:
I just merged into our `soon` branch, which will be in master very soon.

Thanks!

.dockerignore
```diff
@@ -0,0 +1,5 @@
+data
+Dockerfile
+docker-compose.yml
+.git
+.gitignore
```
kpcyrd: please add `node_modules` as well and include `npm install` in the dockerfile

Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM ubuntu:16.04
```
kpcyrd: If you use `FROM node:latest` you get way smaller images because you don't need to run apt-get (also, builds much faster)

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+EXPOSE 3000
+
+VOLUME /cryptpad/datastore
+VOLUME /cryptpad/customize
+
+ENV USE_SSL=false
+ENV STORAGE='./storage/file'
+ENV LOG_TO_STDOUT=true
+
+CMD /container-start.sh
```
kpcyrd: Please include all build steps in the Dockerfile and don't modify the config on start.

Also, use `ENTRYPOINT ["node", "./server.js"]` instead

PR #74, opened 2017-01-12
UI improvements for mobile browsers

brandones:
Features:
- Get rid of margins on mobile
- Disable autozoom on mobile
- Enable toolbar toggling, set to collapsed by default for small screens
- Add a height setting to fix bad 200px default size on mobile (it seems to only affect mobile browsers)

Side effects:
- Disabling autozoom unfortunately disables all zooming on mobile
- Custom height setting is undone by toggling toolbar twice

I think this improves the mobile situation considerably (in a stopgap way until CKEditor 5, which should have real mobile support).

### DIFF
--- a/www/pad/index.html
+++ b/www/pad/index.html
@@ -3,6 +3,7 @@
 <head>
     <title>CryptPad</title>
     <meta content="text/html; charset=utf-8" http-equiv="content-type"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="icon" type="image/png"
         href="/customize/main-favicon.png"
         data-main-favicon="/customize/main-favicon.png"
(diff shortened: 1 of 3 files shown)


brandones:
Looks like tests are failing on Windows due to a Selenium config problem.

cjdelisle:
Thanks for the PR, I'll take a look at this now because we have had a failing test for some time.

cjdelisle:
I fixed the persistent issue with the build, now what broke your PR seems to be something different/transient. https://travis-ci.org/xwiki-labs/cryptpad/builds/191269590

brandones:
It looks like the same error. I rebased onto your master to see if there was new config stuff that needed to be merged in, but that didn't seem to do it either. Any suggestions?

cjdelisle:
Ok it looks like SauceLabs doesn't allow testing pull requests, I just disabled testing on PRs here: https://github.com/xwiki-labs/cryptpad/commit/a8ba798203441dcd1c0c2585b9775d43f07080ee

cjdelisle:
As far as your PR itself, I'm not really qualified to pull it myself, I should let @yflory take a look and get back to you...

yflory:
Merged into our `soon` branch, it will be deployed in the near future.
Thanks a lot for your help!

PR #80, opened 2017-02-03
Customized CSS for Slides and printing CSS

ldubost:
This is not a working solution, but it might be useful in the future for printing.
The main problem is that it does not seem easy to have proper CSS which allows to have a good HTML / CSS for which printing will work.


ansuz:
We've implemented print support in our staging branch. To be merged to master and deployed tomorrow (Tuesday, March 14th), as part of our Chupacabra release.

PR #96, opened 2017-03-03
(CSS) Some improvements to drive

pbondoer:
Various quality of life improvements to /drive 🎉

 * (css) prevent huge titles from making overly large rows - max-height on items
 * (css) prettier overflow with text-overflow: ellipsis
 * (css) add a transparent border when not selected so item size doesnt change and items jump around


ansuz:
LGTM :+1:

PR #105, opened 2017-04-08
Couple of small changes to the docker file

coldfix:
The main issue that caused me to investigate was a startup bug when using the docker image, see first commit. I also found the documentation a bit confusing because the instructions on how to use it were so far from the top.

The last commit is a version update that is (I think) not strictly needed. Include at your will.

Best, Thomas

**EDIT:** And, I almost forgot to mention: Thanks a lof for this wonderful software!! :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,12 @@
-FROM node:6
+FROM node:6-alpine
 
-ARG VERSION=0.3.0
-
-# Download stable version
-# RUN wget https://github.com/xwiki-labs/cryptpad/archive /${VERSION}.tar.gz -O /cryptpad.tar.gz \
-#   && mkdir -p /cryptpad \
-#   && tar -xzf /cryptpad.tar.gz -C /cryptpad --strip-components=1 \
-#   && rm /cryptpad.tar.gz
-
-# Download from github
-# RUN git clone https://github.com/xwiki-labs/cryptpad.git
-
-# Add code directly
-ADD . /cryptpad
-
+COPY . /cryptpad
 WORKDIR /cryptpad
 
-RUN npm install \
+RUN apk add --no-cache git tini \
+   && npm install \
    && npm install -g bower \
    && bower install --allow-root
-
-ADD container-start.sh /container-start.sh
-RUN chmod u+x /container-start.sh
 
 EXPOSE 3000
 
@@ -32,4 +17,4 @@ ENV USE_SSL=false
 ENV STORAGE='./storage/file'
 ENV LOG_TO_STDOUT=true
 
-CMD /container-start.sh
+CMD ["/sbin/tini", "--", "/cryptpad/container-start.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,4 @@
-FROM ubuntu:16.04
-
-RUN apt-get update && apt-get install -y \
-  vim \
-  wget \
-  git \
-  curl \
-  npm \
-  nodejs-legacy
+FROM node:6
 
 ARG VERSION=0.3.0
 
--- a/container-start.sh
+++ b/container-start.sh
@@ -13,7 +13,7 @@ mkdir -p customize
 [[ -n "$USE_SSL" ]] && echo "Using secure websockets: $USE_SSL" \
   && sed -i "s/useSecureWebsockets: .*/useSecureWebsockets: ${USE_SSL},/g" customize/config.js
 
-[[ -n "$USE_SSL" ]] && echo "Using storage adapter: $STORAGE" \
+[[ -n "$STORAGE" ]] && echo "Using storage adapter: $STORAGE" \
   && sed -i "s/storage: .*/storage: ${STORAGE},/g" customize/config.js
 
 [[ -n "$LOG_TO_STDOUT" ]] && echo "Logging to stdout: $LOG_TO_STDOUT" \
--- a/container-start.sh
+++ b/container-start.sh
@@ -1,22 +1,22 @@
-#!/bin/bash
+#!/bin/sh
 
 # Creating customize folder
 mkdir -p customize
-[[ ! "$(ls -A customize)" ]] && echo "Creating customize folder" \
+[ -z "$(ls -A customize)" ] && echo "Creating customize folder" \
   && cp -R customize.dist/* customize/ \
   && cp config.js.dist customize/config.js
 
 # Linking config.js 
-[[ ! -h config.js ]] && echo "Linking config.js" && ln -s customize/config.js config.js
+[ ! -h config.js ] && echo "Linking config.js" && ln -s customize/config.js config.js
 
 # Configure 
-[[ -n "$USE_SSL" ]] && echo "Using secure websockets: $USE_SSL" \
+[ -n "$USE_SSL" ] && echo "Using secure websockets: $USE_SSL" \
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -24,9 +26,9 @@ ADD . /cryptpad
 
 WORKDIR /cryptpad
 
-RUN npm install \
-   && npm install -g bower \
-   && bower install --allow-root
+RUN npm install
+RUN npm install -g bower
```
ansuz: If any one of these instructions fail, then CryptPad won't work. Conjoining these commands with `&&` means it will bail out if anything fails.

container-start.sh
```diff
@@ -20,4 +20,4 @@ mkdir -p customize
   && sed -i "s/logToStdout: .*/logToStdout: ${LOG_TO_STDOUT},/g" customize/config.js
 
 
-exec node ./server.js
\ No newline at end of file
+exec node --use_strict ./server.js
```
ansuz: I'm not sure what version of nodejs ships with Ubuntu these days, but when developing CryptPad we use [NVM](https://github.com/creationix/nvm). Personally I'm using node-v.6.6.0.

If this works, there's nothing to correct, but I'll be able to guarantee that the version I use will continue to work.

ansuz:
Thanks for reviewing this. We don't use docker for [CryptPad.fr](https://cryptpad.fr), or for development, so we don't really keep track of how well it works. It's just here for anyone who wants to use it.

I've linked to this PR in our IRC/Matrix channel, and asked for some feedback from anyone who's using CryptPad via docker. If no one objects, I'll merge this on Monday to be included in Tuesday's release.

ralphtheninja:
@coldfix So currently the docker build is broken right? And it's the first commit that fixes it, so maybe the PR could be narrowed down to this. The `README` contains an error as well (missing the `run`) part so maybe that could also be part of the PR.

coldfix:
Yes. `--use_strict` + the error in the README should definitely be added. The other part is optional. I will rebase if you prefer.

Dockerfile
```diff
@@ -1,14 +1,16 @@
-FROM ubuntu:16.04
+FROM ubuntu:16.10
```
kpcyrd: Consider using https://hub.docker.com/_/node/ instead and removing the `apt-get install` and all the symlinking work

Dockerfile
kpcyrd: `docker build` will bail in both cases, but chaining them with `&&` reduces the number of layers of the final image.

container-start.sh
kpcyrd: Consider using `FROM node:6`

Dockerfile
coldfix: Good suggestion. Thanks! (I don't use any of this technology usually, so I'm happy about feedback from more experienced people!)

coldfix:
Okay. Using @kpcyrd's suggestion now which is a lot better.

kpcyrd:
I've spent some more time with this and came up with this patch. It's based on the current staging branch (sorry), so you might have to apply the changes manually.

- Use `node:6-alpine` as base, this reduces the image size from 638MB (original staging branch) to 156MB
- Replace all `ADD` commands with `COPY`
- Remove the comments, the correct way to do something like this is 1) clone the repo 2) checkout a specific release 3) build image
- Don't copy container-start.sh twice and `chmod +x container-start.sh` in git to save 2 layer
- Use `tini` as minimal init system for signal handling

```
diff --git a/Dockerfile b/Dockerfile
index 99b9150..96a5ff5 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,36 +1,13 @@
-FROM ubuntu:16.04
-
-RUN apt-get update && apt-get install -y \
-  vim \
-  wget \
-  git \
-  curl \
-  npm \
-  nodejs-legacy
-
-ARG VERSION=0.3.0
-
-# Download stable version
-# RUN wget https://github.com/xwiki-labs/cryptpad/archive /${VERSION}.tar.gz -O /cryptpad.tar.gz \
-#   && mkdir -p /cryptpad \
-#   && tar -xzf /cryptpad.tar.gz -C /cryptpad --strip-components=1 \
-#   && rm /cryptpad.tar.gz
-
-# Download from github
-# RUN git clone https://github.com/xwiki-labs/cryptpad.git
-
-# Add code directly
-ADD . /cryptpad
+FROM node:6-alpine
 
+COPY . /cryptpad
 WORKDIR /cryptpad
 
-RUN npm install \
+RUN apk add --no-cache git tini \
+   && npm install \
    && npm install -g bower \
    && bower install --allow-root
 
-ADD container-start.sh /container-start.sh
-RUN chmod u+x /container-start.sh
-
 EXPOSE 3000
 
 VOLUME /cryptpad/datastore
@@ -40,4 +17,4 @@ ENV USE_SSL=false
 ENV STORAGE='./storage/file'
 ENV LOG_TO_STDOUT=true
 
-CMD /container-start.sh
\ No newline at end of file
+CMD ["/sbin/tini", "--", "/cryptpad/container-start.sh"]
diff --git a/container-start.sh b/container-start.sh
old mode 100644
new mode 100755
index db0bb92..662322c
--- a/container-start.sh
+++ b/container-start.sh
@@ -1,23 +1,23 @@
-#!/bin/bash
+#!/bin/sh
 
 # Creating customize folder
 mkdir -p customize
-[[ ! "$(ls -A customize)" ]] && echo "Creating customize folder" \
+[ -z "$(ls -A customize)" ] && echo "Creating customize folder" \
   && cp -R customize.dist/* customize/ \
   && cp config.js.dist customize/config.js
 
 # Linking config.js 
-[[ ! -h config.js ]] && echo "Linking config.js" && ln -s customize/config.js config.js
+[ ! -h config.js ] && echo "Linking config.js" && ln -s customize/config.js config.js
 
 # Configure 
-[[ -n "$USE_SSL" ]] && echo "Using secure websockets: $USE_SSL" \
+[ -n "$USE_SSL" ] && echo "Using secure websockets: $USE_SSL" \
   && sed -i "s/useSecureWebsockets: .*/useSecureWebsockets: ${USE_SSL},/g" customize/config.js
 
-[[ -n "$USE_SSL" ]] && echo "Using storage adapter: $STORAGE" \
+[ -n "$STORAGE" ] && echo "Using storage adapter: $STORAGE" \
   && sed -i "s/storage: .*/storage: ${STORAGE},/g" customize/config.js
 
-[[ -n "$LOG_TO_STDOUT" ]] && echo "Logging to stdout: $LOG_TO_STDOUT" \
+[ -n "$LOG_TO_STDOUT" ] && echo "Logging to stdout: $LOG_TO_STDOUT" \
   && sed -i "s/logToStdout: .*/logToStdout: ${LOG_TO_STDOUT},/g" customize/config.js
 
 
-exec node ./server.js
\ No newline at end of file
+exec node ./server.js
```

coldfix:
Great. I will add this, but since you know more about this than me, you could just do an alternative PR if anything else comes up  :)

Quick question for my understanding: What is the purpose of adding git in the image?

coldfix:
Oh, I see, it's somehow needed for the build process. 

```
...
bower jquery#~2.1.3             ENOGIT git is not installed or not in the PATH
The command '/bin/sh -c apk add --no-cache tini    && npm install    && npm install -g bower    && bower install --allow-root' returned a non-zero code: 1
```

ansuz:
I've merged into our staging branch, and merged that into our soon branch.

Tomorrow morning we'll merge that to master. Thanks everyone!

PR #106, opened 2017-04-09
CSP fix

Aranjedeath:
ws:// is insecure websocket, so browsers will block. fix to use wss:// in CSP config.


Aranjedeath:
closing for a better solution, put both ws:// and wss:// in config, and add note to remove ws:// if deploying over https (basically note that this is development config but not 100% production suitable)

PR #112, opened 2017-04-18
Jquery require2

ansuz:


ansuz:
Accidentally merged to master. I've back-ported it into staging now.

PR #114, opened 2017-05-05
Adding a warning and header to delete/reset docs

realitygaps:
The delete/reset instructions were directly after the upgrade instructions. Figured its useful to put a header and a warning to prevent running those commands by mistake when upgrading.


ansuz:
Thanks!

PR #124, opened 2017-05-27
Creation of config file after running server.js

bashrc:
Without this login and buttons do not appear

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -35,7 +35,9 @@ bower install
 ## copy config.example.js to config.js
 cp config.example.js config.js
 
-node ./server.js
+node ./server.js &
+mkdir -p customize/api
+wget 127.0.0.1:3000/api/config -O customize/api/config
 ```
 
 ## Configuration


ansuz:
Our installation instructions don't account for proxies in front of the nodejs server. The Nginx config file is included as an example for those interested in self-hosting, but for simplicity's sake it isn't the recommended method.

I'll add a note to the top of the Nginx config indicating as much before our next release.

kpcyrd:
> but for simplicity's sake it isn't the recommended method.

You might want to reconsider that, given that https setups for nginx are usually way more complete than nodejs programs attempting to repeat the effort that already went into this.

It might be easier to just avoid situations where it makes a difference if an application is run with or without a reverse proxy.

Ideally, your nginx config would just proxy all requests without modification, do the websocket-upgrade-dance and maybe serve assets directly from disk if you're very serious about performance.

ansuz:
To clarify, our documentation covers how to set up a very basic installation. Essentially it's enough to show that it works, and allow people to test. This is the method we use daily to develop, which doesn't require TLS or high performance.

Setting it up for production use is a different matter, and should probably be in a different document.

PR #129, opened 2017-05-31
XWiki cryptpad updates

Regzox:
Update cryptpad error handling


ansuz:
I've already pushed a built file, but I'm not sure why we got different results from builds. It built from the merged master branch of UCF/media-tag

Regzox:
I've check again and your last media-tag.js push in cryptpad is the same than mine in this PR.
With the last version `NODE_ENV=production-umd yarn build` must build a 54713 bytes bundle.
If is it, versions of cryptpad are same.

PR #131, opened 2017-06-05
Traditional Chinese

jxtsai:
add Traditional Chinese translation


ansuz:
Thanks! There was one typo that prevented `messages.zh.js` from running correctly, but it's fixed. This will be deployed along with tomorrow's release.

PR #139, opened 2017-06-21
add translations and some improvements

dtrv:
Added some missing translations, correct some minor misspellings and try to make the 'zero-knowledge' statement clearer by translating it to 'ohne Preisgabe von Infos' = 'without exposing informations'.


ansuz:
Danke!

PR #142, opened 2017-06-25
Reenable X-Frame-Options: SAMEORIGIN

MacLemon:
This header seems to have been commented out without a reason.
Setting X-Frame-Options to SAMEORIGIN works fine. (Tested in Chromium 59/60/61, Firefox 52/53, Safari 10.1.1)

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -18,7 +18,7 @@ module.exports = {
     httpHeaders: {
         "X-XSS-Protection": "1; mode=block",
         "X-Content-Type-Options": "nosniff",
-        // 'X-Frame-Options': 'SAMEORIGIN',
+        'X-Frame-Options': 'SAMEORIGIN',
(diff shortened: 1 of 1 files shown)


cjdelisle:
We do some remote authentication using an iframe from accounts.cryptpad.fr (so we can auth you if you make a donation) so this header will need to at least include accounts for the /auth/ line but other than that LGTM.

MacLemon:
The legacy of your own installation does not apply to all other installations, especially not new ones. The default should be sane and secure for everyone.
It would be easy for you to replace that specific header in nginx your own deployment to match your authentication legacy and still deliver a secure default for everyone else.

ansuz:
We're dropping support for this legacy issue in tomorrow's release, so SAME-ORIGIN will be enabled again.

Issue #151, opened 2017-07-07
Spacing between "Log-in" and "Register" buttons


evalica:
See Before
<img width="1013" alt="before" src="https://user-images.githubusercontent.com/629552/27960995-edcce218-6335-11e7-9473-417397603114.png">
and After
<img width="1055" alt="after" src="https://user-images.githubusercontent.com/629552/27961011-fa12914e-6335-11e7-8bdc-abbd3edc2415.png">

PR #152, opened 2017-07-07
Issue #151: Spacing between "Log-in" and "Register" buttons

evalica:
- Removed ".first" class from Register button
- Increase spacing between buttons


ansuz:
Thanks!

PR #157, opened 2017-07-20
File size reduction and optimization of new image assets

MacLemon:
Optimized new image asses to speed up page loading and reduce traffic.

screen shot 2017-07-20 at 20 01 03


ansuz:
Thanks again!

PR #158, opened 2017-07-20
Translated missing out.presentSuccess to german

MacLemon:
out.presentSuccess wasn't translated to german yet, added german translation. (I'm a native speaker of german.)


ansuz:
Thanks!

PR #159, opened 2017-07-23
Fixed spelling issues and improved translation quality.

c4rl0sc:
The Spanish translation had several issues, mostly related to accent marks and spelling.

### DIFF
--- a/customize.dist/translations/messages.es.js
+++ b/customize.dist/translations/messages.es.js
@@ -18,7 +18,7 @@ define(function () {
     out.synchronizing = "Sincronización";
     out.reconnecting = "Reconectando...";
     out.lag = "Retraso";
-    out.readonly = 'Solo lectura';
+    out.readonly = 'Sólo lectura';
     out.anonymous = 'Anónimo';
     out.yourself = "Tú mismo";
     out.anonymousUsers = "usuarios anónimos";
@@ -30,8 +30,8 @@ define(function () {
     out.editor = "editor";
     out.editors = "editores";
 
-    out.greenLight = "Todo funciona bién";
-    out.orangeLight = "La conexión es lenta y podria impactar la experiencia";
+    out.greenLight = "Todo funciona bien";
+    out.orangeLight = "La conexión es lenta y podría afectar la experiencia";
     out.redLight = "Has sido desconectado de la sesión";
 
     out.importButtonTitle = 'Importar un documento de tus archivos locales';
@@ -44,7 +44,7 @@ define(function () {
     out.clickToEdit = "Haz clic para cambiar";
 
     out.forgetButtonTitle = 'Eliminar este documento de la lista en la pagina de inicio';
-    out.forgetPrompt = 'Pulser OK eliminará este documento del almacenamiento local (localStorage), ¿estás seguro?';
+    out.forgetPrompt = 'Pulsar OK eliminará este documento del almacenamiento local (localStorage), ¿estás seguro?';
 
     out.shareButton = 'Compartir';
     out.shareSuccess = 'URL copiada al portapapeles';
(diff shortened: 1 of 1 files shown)


ansuz:
@pbondoer can you review?

customize.dist/translations/messages.es.js
```diff
@@ -18,7 +18,7 @@ define(function () {
     out.synchronizing = "Sincronización";
     out.reconnecting = "Reconectando...";
     out.lag = "Retraso";
-    out.readonly = 'Solo lectura';
+    out.readonly = 'Sólo lectura';
```
pbondoer: I believe this is incorrect: http://www.rae.es/consultas/el-adverbio-solo-y-los-pronombres-demostrativos-sin-tilde

customize.dist/translations/messages.es.js
```diff
@@ -337,25 +337,25 @@ define(function () {
...
     out.settings_import = "Importar";
-    out.settings_importConfirm = "¿Seguro qué quieres importar tus pads recientes a tu cuenta CryptDrive?";
+    out.settings_importConfirm = "¿Seguro que quieres importar tus pads recientes a tu cuenta CryptDrive?";
     out.settings_importDone = "Importación terminada";
 
     out.tips = {};
     out.tips.lag = "El icono verde en la parte superior derecha muestra la calidad de tu connexión a CryptPad.";
     out.tips.shortcuts = "`ctrl+b`, `ctrl+i`, y `ctrl+u` son accesos rápidos para negrita, itálica y subrayado.";
-    out.tips.indent = "Cuando editas listas, puedes usar tab o shift+tab para icrementar o decrementar indentación.";
+    out.tips.indent = "Cuando editas listas, puedes usar tab o shift+tab para incrementar o decrementar la sangría.";
```
pbondoer: I've never seen the word "sangría" used in this context, perhaps "margen" would be a better fit?

customize.dist/translations/messages.es.js
gustavohmsilva: The ó and o difference is provenient of the difference between "es" and "es_ES" languages.

customize.dist/translations/messages.es.js
gustavohmsilva: You are correct, "sangría" is related to diagramation design terminology, "margen" is related to text editing terminology

customize.dist/translations/messages.es.js
pbondoer: Do other variants of Spanish have their own accentuation rules? I thought they would follow what the Spain variant does.

c4rl0sc: Thank you for reviewing.

You're right, the Spanish language authority (RAE) deprecated the use of "sólo" a few years ago. It's funny because I trusted my Windows 7 installation, which reads "Sólo lectura", but the truth is that Windows 10 has already updated this piece of text with the new rule, so that it now reads "Solo lectura".

I don't know whether RAE rules are applicable to other Spanish variants different from es_ES.

To sum up: I think that the accent should be removed.

customize.dist/translations/messages.es.js
c4rl0sc: I trusted my Microsoft Word and LibreOffice Writer installations, which have both translated "identation" as "sangría", apparently in the same context as in CryptPad. To be honest, I've never heard the term "diagramation design". On the other hand, "margen" has always sounded more natural to me. So I don't know what to say.

pbondoer: If this is the term LibreOffice and Word use, then it's OK for me 🎉

customize.dist/translations/messages.es.js
pbondoer: This translation was intended as `es_ES`. Language variants are a bit difficult to deal with, since it's not always clear what the differences are, and making a culture invariant version of a translation is, in my experience, almost impossible. The compromise I've seen is to either not offer invariants, or to default the invariant to the most common variant. Take the case of `en_US`/`en_GB`: how do you choose between US spellings (color) or GB spellings (colour)? 🤔

customize.dist/translations/messages.es.js
gustavohmsilva: c4rl0sc. Sorry, it was an error, is only diagramation :D

customize.dist/translations/messages.es.js
gustavohmsilva: pbondoer:  ~~almost~~ impossible. I work on a company who makes internationalization on a daily basis (but mainly for Subtitles, Transcripts, etc). The rule **here** is going for the variation of the language with the bigger population worldwide. I'm saying this just as a point of comparison.

customize.dist/translations/messages.es.js
```diff
@@ -397,9 +397,9 @@ define(function () {
     out.history_closeTitle = "Cerrar el historial";
     out.history_restore = "Restaurar";
     out.history_restoreTitle = "Restaurar la versión seleccionada del documento";
-    out.history_restorePrompt = "¿Estás seguro que quieres cambiar la versión actual del documento por esta?";
+    out.history_restorePrompt = "¿Estás seguro de que quieres cambiar la versión actual del documento por ésta?";
```
c4rl0sc: If you want to follow the new rules by RAE (applicable to `es_ES`, at least), "ésta" should be reverted back to "esta".

Source: http://www.rae.es/consultas/el-adverbio-solo-y-los-pronombres-demostrativos-sin-tilde

ansuz:
Gracias!

PR #166, opened 2017-08-09
Properly escape sed match/replace expressions

coldfix:
This fixes an error during container startup due to interpolating a
$STORAGE value that may contain slashes or other "active" characters:

    ...
    Using secure websockets: false
    Using storage adapter: ./storage/file
    sed: bad option in substitution expression

I guess this is the same as the TODO item mentioned in the
docs/cryptpad-docker page.

Also add a trailing ',' in the search expression to avoid substituting
in the comment above the actual definition:

    /*  If Cryptpad is proxied without using https, the server needs to know.
     *  Specify 'useSecureWebsockets: true' so that it can send
    ...

By the way, I have a related question: I'm using cryptpad behind an nginx reverse proxy (SSL). Should I specify `useSecureWebsockets: true` or false? I'm a bit confused by the wording.

Best, Thomas

### DIFF
--- a/container-start.sh
+++ b/container-start.sh
@@ -9,15 +9,20 @@ mkdir -p customize
 # Linking config.js
 [ ! -h config.js ] && echo "Linking config.js" && ln -s customize/config.js config.js
 
+# Thanks to http://stackoverflow.com/a/10467453
+sedeasy() {
+  sed -i "s/$1/$(echo $2 | sed -e 's/[\/&]/\\&/g')/g" $3
+}
+
 # Configure
 [ -n "$USE_SSL" ] && echo "Using secure websockets: $USE_SSL" \
-  && sed -i "s/useSecureWebsockets: .*/useSecureWebsockets: ${USE_SSL},/g" customize/config.js
+  && sedeasy "useSecureWebsockets: [^,]*," "useSecureWebsockets: ${USE_SSL}," customize/config.js
 
 [ -n "$STORAGE" ] && echo "Using storage adapter: $STORAGE" \
-  && sed -i "s/storage: .*/storage: ${STORAGE},/g" customize/config.js
+  && sedeasy "storage: [^,]*," "storage: ${STORAGE}," customize/config.js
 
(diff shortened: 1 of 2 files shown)


ansuz:
I believe that text in the config is there from some legacy configurations. We use `websocketPath` on CryptPad.fr.

The relevant function for fetching the websocket URL (clientside) is here: [getWebsocketURL()](https://github.com/xwiki-labs/cryptpad/blob/master/www/common/cryptpad-common.js#L277-L287)

This function reads values from `/api/config`, which is generated in the server [here](https://github.com/xwiki-labs/cryptpad/blob/master/server.js#L124-L139).

If `useExternalWebsocket` is falsey, then clients will use websocketPath instead of websocketURL.

For your Nginx proxy with SSL, clients should pull from /api/config and infer the correct location of the websocket in one of two ways:

1. If you use the default relative path from the config (`/cryptpad_websocket`), clients will infer the absolute path using the correct protocol (`wss://` if they have connected over `https://`), using the same domain.
2. if you provide an absolute path (ie. on a different domain) clients will connect to that exact path.

---

It's probably safe to simply remove a few of these confusing parameters from the config, or at least mark them as deprecated.

coldfix:
Ok, thanks! So `useSecureWebsockets` is only ever used if `useExternalWebsocket` is truish. I agree with the implementation that having the client determine the websocketPath based on the protocol/port in use is usually the most reliable way.

I think it would make sense to remove at least the `USE_SSL` parameter from the docker file as it will by default have no influence unless also modifying the config file. And once one starts using their own config, it doesn't really make sense to have it modified based on environment variables. In fact, I'd probably also remove the other parameters config parameters from the docker file and always rely on the config file instead.

In fact, to my understanding, I would completely remove the `websocketPort` and `useSecureWebsocket` variables from the config; let the client always use `websocketPath` from the config and use `useExternalWebsocket` only to check whether the server should start a server (is that setting ever used btw?).

I'll assemble a PR.

PR #167, opened 2017-08-10
Remove useSecureWebsockets variable, further simplify docker image

coldfix:
following discussion in #166.

I found the useSecureWebsockets variable rather confusing, and as far as I can see also not useful anymore in the current implementation. Removing it and some of the other env vars help it KISS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,4 @@ EXPOSE 3000
 VOLUME /cryptpad/datastore
 VOLUME /cryptpad/customize
 
-ENV STORAGE='./storage/file'
-ENV LOG_TO_STDOUT=true
-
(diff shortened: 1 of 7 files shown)


ansuz:
There shouldn't be any changes to the server here, as they would break CryptPad for people still using the old configuration values. Changing the example configuration file only will make it more straightforward for people setting up fresh installations, however. The same is true of the line in `cryptpad-common.js`.

coldfix:
Ok, I can understand that you want to keep backward compatibility, even though I feel like the behaviour of the option was rather surprising if not counterproductive: if a separate socket server is used is one of the only cases where you want to set *websocketPath* explicitly in the config file and only in this case it is not possible to do directly.

PR #170, opened 2017-08-17
Use CryptPad everywhere instead of Cryptpad in titles

pbondoer:
Noticed there were some residual "Cryptpad" titles, so I changed the casing on those, as it would flash briefly in the title bar 📸


ansuz:
thanks!

PR #181, opened 2017-09-18
Cryptad version in Readme.md added

genma:
Because this information was lacking and I need to find it quickly (and probably other people).
Don't forget to update this part at each new released.


ansuz:
I think the information you are looking for is [here](https://github.com/xwiki-labs/cryptpad/releases).

genma:
Yes. A link to this page in the Read.me is a good pratice / thing in my point of view. Thanks for the link (so I can propose to add it).

PR #185, opened 2017-10-13
Compressed images for faster connections

MacLemon:
Recompressed new image assets to speed up connections and reduce data to be transferred.

PNGs were recompressed pixel identically.
JPGs were recompressed visually identically.

Saving about 9.8%≈580kB overall, to speed up client connections.


ansuz:
Thanks!

PR #190, opened 2017-10-26
Theming improvements as proposed from dyne

evilham:
We have discussed this over IRC and email, I think this commit improves the ability to theme cryptpad by moving some colours to colortheme.less under customize.dist.

### DIFF
--- a/customize.dist/src/less2/include/colortheme.less
+++ b/customize.dist/src/less2/include/colortheme.less
@@ -41,6 +41,7 @@
 
 @colortheme_pad-bg: #1c4fa0;
 @colortheme_pad-color: #fff;
+@colortheme_pad-toolbar-bg: #c1e7ff;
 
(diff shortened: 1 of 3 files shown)


ansuz:
I just tested this locally, and it looks like some new variables are used which only exist in your patched colortheme, or something similar. As a result, the undefined variable causes the less compiler to abort, which breaks all of our styles.

I'm ok to merge once it a) works without any patches in /customize and b) preserves our existing color theme. The goal being to make it easier for others to apply skin, should they wish.

evilham:
Of course, how didn't I think of that? :)

I *think* my last commit fixes that. Do notice I'm pretty much a LESS noob (nor do I currently want to become an expert on that), so please don't hesitate telling if it should be done in a different fashion (and please how :)).

ansuz:
Thanks!

PR #213, opened 2017-12-24
Update example.nginx.conf

Falgn0n:
Updated the location block for "/api/config" to point directly to the node.js server @ localhost:3000


ansuz:
Thanks!

PR #222, opened 2018-01-23
Fix docker build

kpcyrd:
This fixes the `Failed at the heapdump@0.3.9 install script 'node-gyp rebuild'.` issue so docker builds again.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,7 @@
 data
 Dockerfile
 docker-compose.yml
+.dockerignore
 .git
 .gitignore
+node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,4 @@ node_js:
 script:
   - npm run-script lint
   - npm run-script flow
+  - docker build -t xwiki/cryptpad .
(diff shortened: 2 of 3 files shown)


ansuz:
We don't use docker ourselves, so I have no issues merging this if it suits your use case.

I wonder, however, if you'll be able to use the heap dump functionality while the app is contained in docker. It's there for debugging purposes, and if it will not be used, a simpler solution would be for us to comment out the heapdump require call in `config.example.js`.

kpcyrd:
Technically you could `docker exec` into the container if needed, but getting rid of heapdump would work as well as it's currently listed as a dev dependency. I've started using cryptpad with docker recently, please feel free to ping me on github if there are any  docker related issues you need reviewed. :)

ansuz:
I'm going to comment out heapdump from the example config file.

Dev dependencies are skipped if you install with `npm install --production`, so if the docker configuration does that, I believe everything should work again.

kpcyrd:
@ansuz thanks, I've updated my PR.

PR #227, opened 2018-02-09
Fixed customization of main pages

nparquet:
It appears that even though terms.html, index.html etc ("main pages") are in the customize.dist folder, we could not customize these pages by creating a similarly-named file in the customize folder.

Adding this line reuses the same mecanism as lines 127-128 to recover this behavior on main pages.


ansuz:
Thanks!

PR #229, opened 2018-02-14
add license to package.json

neuhaus:
the project uses AGPL-3.0


ansuz:
Hi,
thanks for working on this.

I see that you've listed it as the AGPL-3.0-only, in contrast to allowing for later versions of the AGPL.

The choice of license predates my contributions to the project, and I'm not aware of where the LICENSE text was sourced from. Is there a tangible difference between the text of the two, or did you choose `-only` just to be more exact?

ansuz:
I compared the two as found on the [SPDX license list](https://spdx.org/licenses/), and they seem to have the same text, but _I am not a lawyer_, and so I prefer to be careful about things like this.

ansuz:
Our readme specifies that we accept a later version

> either version 3 of the License, or (at your option) any later version

neuhaus:
thanks! I updated the patch.

PR #240, opened 2018-04-26
Build on openshift

kanedafromparis:
This is a small patch to allow it to be build in openshift.

Fell free to give me feddback ;-)


ansuz:
We have a number of scripts that include the configuration file in its current location. I realize we haven't written tests for those, so it's not immediately clear that those are required to work.

With that being said, as Openshift seems to be mostly a platform for managing Docker images, I'm curious why our existing docker setup is not sufficient.

kanedafromparis:
So, there is 2 points.

1 - in order to have config.js file external to your project (and specially not in the git)
you need to mount your config file into a directory 

https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#add-configmap-data-to-a-volume
( I say openshift but it's a kubernetes way)


2 - Openshift build system
you have 2 way to build you image using Dockerfile, or [s2i](https://github.com/sclorg/s2i-nodejs-container/tree/master/6), if you use s2i, your patch management (aka security update) is handle by your RedHat subscription (which is easier on the long run) so in order to have a build process that works I add "post-install script" is needed

Issue #169, opened 2017-08-15
Anchors should output... anchors (or IDs)


polx:
The feature with the "flag" icon seems just right: I can input an anchor name and I can potentially link to that. The problem is the potentially: What is the way for me to link to that?

Neither HTML exports, nor the read-only view display this anchor as an <a name="..."> or id attribute.

I must have overlooked something.

ansuz:
I've gone ahead and removed the anchor button as it doesn't seem especially useful.

polx:
Mmmh. I don't think that this is a good idea, except as a temporary idea.
If you tell me where I can hunt, I am happy to hunt.

For me this feature is part of a bigger set of features where CryptPad documents are the editing side of the world and the HTML export is used to produce, say, a website with a generator, e.g. Jekyll.

paul

ansuz:
This looks like a broken feature buried somewhere in CKEditor. It's possible to create a link to the anchor with the link button, however, the anchors don't seem to produce anything usable by browsers.

We could open an issue on CKEditor's issue tracker, however, I think they're putting all of their effort into CKEditor 5, so this would probably be fruitless.

polx:
Isn't it a matter exchanging the output of CKEditor?
Saving the html that CKEditor manipulates clearly is not a way to expect CKEditor to deliver an HTML that is for the world outside.

The following Regexp seems to work for me:
`            html = html.replace(/<img[^>]*class="cke_anchor"[^>]*data-cke-realelement="([^"]*)"[^>]*>/, function(match,realElt){ return unescape(realElt); }); `

But, of course, this bases on a way CKEditor encodes its things.
It wouldn't be very sustainable if this is not counter-checked by some testing.

paul

polx:
Hello Aaron and all,
following this discussion, I created the pull request #245 where the Anchor function is restored, the export indeed yields an <a id='...'> element and a test corresponds to this.

Note however, that this requires the tested /pad/ documents to contain at least an anchor, as I have found no way in the current test systems to predefine the documents being tested or to press on buttons to do so. Maybe I have missed something...

This manual anchor function should follow the work done for the issue https://github.com/xwiki-labs/cryptpad/issues/243 .

paul

polx:
I believe that my last commit (0dc1500784051226a6bf464b08035ade66958a8d) solves this issue and I propose to merge this.
thanks
paul

polx:
@ansuz, are you willing to review that?
thanks
paul

PR #245, opened 2018-05-11
#169: First anchor test and restore of the CKEditor anchor function.

polx:


ansuz:
It looks like our linter caught some undefined references

```
www/pad/inner.js: line 169, col 49, 'unescape' is not defined.
www/pad/inner.js: line 773, col 35, 'CKEDITOR' is not defined.
www/pad/inner.js: line 778, col 48, 'inner' is not defined.
```

Those may actually be undefined or out of scope, or you may just have to use `window.CKEDITOR`.

ansuz:
I modified a few lines to make it pass our linter, and now the tests pass.

I'll need to look over the changes in more detail to make sure nothing breaks as a result. I'm most concerned about correctly reimporting content that was modified at export time.

polx:
Thanks.
So... That looks like an extra (interesting) requirement.
Can you detail that in the issue maybe?
Is the fact that the data-attribute be (un)changed independently of the anchor a danger?

thanks

paul

ansuz:
I don't know whether it's a danger, just that CKEditor is tricky sometimes. I did an export and import with some anchors, and if something is wrong, I can't tell.

polx:
The reason I am asking, is that it seems like yet another step in test-automation.
paul

polx:
Funnily, I saw recently that the read-only version also does not use the anchor (only the flag picture is there but no id or so; no scrolling that I could see).

PR #253, opened 2018-07-11
Total revision of german translation

ilu33:
Corrected lots of typos, grammar and incomprehensible (machine?) translation. Things can still be improved but at least most of the strangeness should be gone. I also added all strings from messages.js that were not included before.

Somebody needs to check that I did not mess up code/formatting.

### DIFF
--- a/customize.dist/translations/messages.de.js
+++ b/customize.dist/translations/messages.de.js
@@ -1,17 +1,18 @@
-  define(function () {
-      var out = {};
+define(function () {
+    var out = {};
 
-      // translations must set this key for their language to be available in
-      // the language dropdowns that are shown throughout Cryptpad's interface
-      out._languageName = 'German';
+    // translations must set this key for their language to be available in
+    // the language dropdowns that are shown throughout Cryptpad's interface
+    out._languageName = 'German';
 
-      out.main_title = "Cryptpad: Echtzeitzusammenarbeit ohne Preisgabe von Informationen";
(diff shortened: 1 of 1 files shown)


polx:
Hello @ilu33,

thanks for the correction. The der/die/das adjustments are certainly welcome and several other changes show that you are probably a native speaker which I am not.

The biggest row of changes you made, however, touches the delicate choice of address which we debated sometimes ago already: Whether cryptpad should use "du" or "Sie"; your changes suggest to use "Du" (and "Dein" and... all capital).
As far as I know this is yet another rather traditional style German, but not as traditional as "Sie".

Can you substantiate this a bit? I do not see a big motivation for that.
Maybe it is useful to ask @michaelachmann and @kpcyrd or simply refer your pull-request on https://github.com/xwiki-labs/cryptpad/issues/239 .

paul

ansuz:
As @polx pointed out, there was a considerable amount of discussion before we settled on how to address the user.

As I only speak Deutsch well enough to order a beer and currywurst with fries, I'd prefer to hear from more users before merging or requesting revisions.

kpcyrd:
I have to admit I didn't know the uppercase informal "Du" in the middle of a sentence exists. The formal "sie" is also always lowercase unless you're writing a formal letter/email. Keeping it lowercase might be a good idea.

polx:
@ilu33: Several sites use "du" with lowercase. E.g. Ikea, Facebook, AirBnB, and to me it's bothering but I am old too! The _personal_ "Du" is something I had only seen in personal letters or emails thus far and this is the kind of hints that the Duden page gives.
paul

ilu33:
Please go ahead.

ansuz:
thanks !

Issue #244, opened 2018-05-09
Cryptpad with docker: Adding volume for blob/... causes EXDEV-error


SPQRInc:
Hello everybody,

I added these volumes to `docker-compose.yml`

```
    volumes:
      - ./data/files:/cryptpad/datastore:rw
      - ./data/customize:/cryptpad/customize:rw
      - ./data/blob:/cryptpad/blob:rw
      - ./data/blobstage:/cryptpad/blobstage:rw
      - ./data/pins:/cryptpad/pins:rw
```

After building my container cryptpad works fine - uploading a file is possible (100%) but once the file is being moved from `blobstage` to `blob` I'm getting this error:

```
2018-05-09T11:08:00.601Z [rename] { Error: EXDEV: cross-device link not permitted, rename 'blobstage/4A/4A2WCNiR2lKSg9ANT2dT-i8SRGjMyXbxVl36oiujHuE=' -> 'blob/f4/f4cc4c20cd69149a1016932fea0c5f4955873bbb10f01bb4'
    at Error (native)
```

How can I solve this?

mcrosson:
This is a bug within aufs by the looks of other projects having similar issues reported.

It looks like the fix may be to use move instead of rename functions.

mcrosson:
The below diff will resolve the issue with avatars

```
diff --git a/rpc.js b/rpc.js
index 69d84b6..b13d1c9 100644
--- a/rpc.js
+++ b/rpc.js
@@ -7,6 +7,8 @@ var Nacl = require("tweetnacl");
 /* globals process */

 var Fs = require("fs");
+Fs.move = require('fs-extra').move;
+Fs.moveSync = require('fs-extra').moveSync;
 var Path = require("path");
 var Https = require("https");
 const Package = require('./package.json');
@@ -1054,7 +1056,7 @@ var upload_complete = function (Env, publicKey, id, cb) {
         }

         // lol wut handle ur errors
-        Fs.rename(oldPath, newPath, function (e) {
+        Fs.move(oldPath, newPath, function (e) {
             if (e) {
                 WARN('rename', e);
                 return void cb('RENAME_ERR');
@@ -1146,7 +1148,7 @@ var owned_upload_complete = function (Env, safeKey, cb) {

         // flow is dumb and I need to guard against this which will never happen
         // / *:: if (typeof(oldPath) === 'object') { throw new Error('should never happen'); } * /
-        Fs.rename(oldPath, finalPath, w(function (e) {
[...]

PR #256, opened 2018-07-16
Fixup docker rename calls (addresses #244)

mcrosson:
The aufs storage driver in docker will not work with the rename Fs call. This patch updates to use the move Fs call instead which works with the aufs Docker storage driver.

### DIFF
--- a/rpc.js
+++ b/rpc.js
@@ -7,6 +7,8 @@ var Nacl = require("tweetnacl");
 /* globals process */
 
 var Fs = require("fs");
+Fs.move = require('fs-extra').move;
+Fs.moveSync = require('fs-extra').moveSync;
 var Path = require("path");
 var Https = require("https");
 const Package = require('./package.json');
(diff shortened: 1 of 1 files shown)


kpcyrd:
The aufs storage driver is known to have issues and not part of the kernel, but a 3rd party kernel module. Please migrate to eg. [overlayfs] instead.

[overlayfs]: https://sthbrx.github.io/blog/2015/10/30/docker-just-stop-using-aufs/

mcrosson:
This also applies to overlay2 as the driver I just re-checked the configuration and the behavior was present with overlay2 as well.

```
lollipopcloud ~ # docker info | grep -i storage
Storage Driver: overlay2
WARNING: No swap limit support
```

ansuz:
All of your other PRs only touched files which will not affect the deployment we're planning for later in the day. I'll have to spend a bit more time reviewing this before pulling.

Having taken a quick look at how fs-extra implements `move`, it seems fairly safe. I'll see if I can adjust the changes to `rpc.js` so as to make them pass flow's lint process.

mcrosson:
@ansuz Sounds good :+1:

ansuz:
I've merged this fix into our staging branch, with some minor adjustments so that it passes our lint process.

mcrosson:
Should this PR be closed now that the update is on the staging branch?

ansuz:
@mcrosson the PR will be automatically marked as merged once I've merged staging to master. I'll be working on that later today.

PR #257, opened 2018-07-16
Update docker start script to bust the cache on each run

mcrosson:
Update docker run script to bust the cache on each run. This is to ensure there aren't left over cache items when a container is started with a new version or the config has been changed heavily.


ansuz:
Thanks!

PR #258, opened 2018-07-16
Add Dockerfile for arm64 support

mcrosson:
Cryptpad will run on arm64 but needs a totally different base image and some other Dockerfile tweaks. This PR adds a Dockerfile.arm64 that can be used to deploy on arm64 boards/VPS instances like the Raspberry Pi or the Scaleway arm64v8 offerings.


ansuz:
Thanks again!

PR #268, opened 2018-08-14
Update readme.md

Wikinaut:
s/ChainPad/CryptPad/


ansuz:
The current wording refers to [ChainPad](https://github.com/xwiki-contrib/chainpad), one of CryptPad's core libraries. That being said, people shouldn't have to know about all our libraries in order to reason about CryptPad's behaviour, so I'm fine to merge anyway.

PR #271, opened 2018-08-28
[WIP] Partially Fix #254

### LINKED ISSUE #254: Kanban feature wishlist
- Export into xml/json
- 3 Item properties: Title, assignee and a per default hidden field for details/remarks/links
- background colors for items, color picker with #000000 color codes
- more colors for boards, color picker with #000000 color codes
- more space on the board - too much wasted whitespace at the moment 
- (option for a) condensed layout
- move the + sign into the toolbar
- resizable font size or at least much smaller defaults

MTRNord:
Hi :) This is the PR that got mentioned in https://github.com/xwiki-labs/cryptpad/issues/254#issuecomment-409908429

In the minute I am writing this this proapbly is outdated. This will be fixed in a bit. (I made this a month ago)

Features this adds are: 
- Adding a color picker for board titles (it still uses default colors)
- Make the huge space between the boards smaller
- Adding a color picker to the items.

PS: Some changes that include are whitespaces. I am sorry for them. My IDE (idea webstorm) refused to not reformat it. :/

Also I do agree to the CLA but the page doesnt let me and instead promts "We're sorry, but something went wrong."


MTRNord:
I dont know who is in charge but I dont think the thord party libary used in here should fail the tests but should be excluded. Or did I miss something?

ansuz:
> Am I correct that jscolor.js needs to be added to the jshintingnore?

That's correct.

www/kanban/app-kanban.less
```diff
@@ -117,39 +117,39 @@
         }
 
         .kanban-header-yellow {
-            background: #FC3;
+            background: #FC3 !important;
```
ansuz: why is `!important` necessary here ? We do our best to avoid any usage of this.

www/kanban/inner.js
```diff
@@ -211,22 +212,26 @@ define([
...
-                verbose("Next color " + nextColor);
-                boardJSON.color = nextColor;
-                $(el).removeClass("kanban-header-" + currentColor);
-                $(el).addClass("kanban-header-" + nextColor);
-                kanban.onChange();
+                jscolorL.fromString(currentColor);
             },
-            buttonClick: function (el, boardId, e) {
-                e.stopPropagation();
+            buttonClick: function (el, boardId) {
```
ansuz: what are the implications of removing the `stopPropagation` call? I don't see how that relates to the color picker.

www/kanban/jkanban.js
```diff
@@ -144,15 +145,18 @@
...
-                                moves: function (el, source, handle, sibling) {
-                                    if (self.options.readOnly) { return false; }
-                                    return handle.classList.contains('kanban-item');
-                                },
-                                accepts: function (el, target, source, sibling) {
-                                    if (self.options.readOnly) { return false; }
-                                    return true;
-                                },
-                                revertOnSpill: true
+                            moves: function (el, source, handle, sibling) {
```
ansuz: this looks like mostly changes to whitespace. Is it possible to minimize the changes to functional changes.

www/kanban/inner.js
MTRNord: I dont think I did that intentional. I will recheck and readd if it hs no change to what I added

www/kanban/jkanban.js
MTRNord: I will try to convince my IDE about that

www/kanban/app-kanban.less
MTRNord: This is because jscolor would otherwise overwrite the default colors with white. It is simply to make sure that default colors have a higher initial priority than jscolor has as no color is yet picked.

MTRNord:
I will work on the comments on the next weekend.

www/kanban/inner.js
```diff
@@ -10,6 +10,7 @@ define([
     '/common/modes.js',
     '/customize/messages.js',
     '/kanban/jkanban.js',
+    '/kanban/jscolor.js',
```
ansuz: I'm curious, what does jscolor do that the native colorpickers can't? I hadn't realized that this PR pulled in a library until now, and we'll have to do a bit of a review of its code before pulling it in.

MTRNord: The most important thing it can do instead of native color pickers is that (atleast on windows) the native color picker is a popup window. Which in my opinion does a better workflow in an app like this. (A popup feels more disctracting than this "overlay popup")

Their page is here: http://jscolor.com/

ansuz: Thanks! I agree about the overlay being less distracting. I'll put some time towards reading through their source, but as long as there's nothing that poses a security risk, I'm in favour of pulling it in.

ansuz:
Thanks!

Issue #270, opened 2018-08-17
Don't automatically add open tabs with cryptpads to the drive


b3yond:
## Expected Behaviour

When I log into a commonly used cryptpad account, but have private cryptpads open, I don't want them automatically added to the drive. Instead it could ask me in each tab "do you want to add this cryptpad to your account?" This way, only those cryptpad links are exposed, which I want to share.

## Actual Behaviour

As soon as I log into a cryptpad account, all open cryptpads are silently added to it. If you don't expect this, you won't even realize (maybe even for months), exposing your private or possibly incriminating information to others. 

## Steps to Reproduce

1. create a cryptpad and fill it with private data (be creative, maybe a diary with all the times you cheated your wife or stole food in a supermarket?)
2. log in to your family's cryptpad (e.g. to plan your common vacations)
3. optional: be disinherited

## Why do I think this should be the default behaviour?

A friend of mine realized recently, that they leaked legally incriminating information to a larger group, because they didn't know cryptpad's default behaviour. it was not discovered for months.

This is the definition of a data breach through human error. Unfortunately, we can't expect everyone to understand how the software works - in many times, they just hope that it keeps them safe.

(and then they ask me to open a bug report for them, because that's what nerd friends are good for, apparently)

I love cryptpad and I use it a lot - for private matters *and* for group planning and collaboration. This is why this is so hazardous. Of course you could easily circumvent this by just using private tabs for every new account or whatever - but people just don't expect this.

I hate that people rely on software, and stop thinking for themselves. I give workshops about technology, so people start getting into it, and I could also just mention this behaviour in every cryptoparty I organize. But this is not scalable. So please design your software for stupid users, this also lowers the barrier, and enables more people to use it and protect themselves.

Thanks for this wonderful service, btw!

ansuz:
Thanks for the detailed report, this is very valuable feedback!

As someone who knows that pads are saved implicitly, I've come to rely on that behaviour. With that being said, I recognize that we don't communicate this behaviour anywhere in the UI, and I appreciate the risk involved for anyone sharing computers or accounts with other users.

It should be possible to support both behaviours in a visible but non-intrusive manner, but we'll need to put some thought into how to best accommodate either workflow.

PS: I hope your friend was not disinherited!

b3yond:
Thanks :) This sounds reasonable.

This behaviour could already be enough:

* When you login with a cryptpad account, 
  * all pads which are opened in other windows 
  * AND which are not in the account yet
[...]

PR #273, opened 2018-09-06
Changed the description for autostore settings (en/de)

b3yond:
Hi, I didn't ilke the strings for the autostore settings (#270).

I changed them and added the german translations for the whole "choose autostore" feature.

I hope this suffices, do you need me to test them in the webUI or something similar?


ansuz:
I can't comment on the quality of the German translations, but since they're insertions I don't see any problem merging.

I agree with the improvements to the English version.

Thanks for the help!

PR #274, opened 2018-09-11
Update messages.de.js: typo: Markdwon -> Markdown

jedie:
typo


polx:
Thank you Jedie,

good catch.
Btw, if you are interested to the German translation of CryptPad, you may be interested to comment on the [issue 255](https://github.com/xwiki-labs/cryptpad/issues/255) about wether we should use "du", "Sie" or "Du".
thanks again.
paul

Issue #198, opened 2017-11-14
🖨 Feature request: print Markdown in /code


pbondoer:
I often write up my documents in Markdown using the `/code` editor. However sometimes I would like to send a pretty formatted document. It would be nice to be able to print the rendered Markdown, so that we can use the browser's "Print to PDF" feature, similarly to what was done for `/slide`.

This might require an `<iframe>`, though I'm not entirely sure as I haven't touched the app in a while. I could try to free up some time to make a PR for this if you are busy working on other features 🙂

PR #302, opened 2018-12-10
Issue #198: Add printing option to /code

joldie:
Hi,

This adds print functionality to the `/code` pad, as per feature request [#198](https://github.com/xwiki-labs/cryptpad/issues/198).

It's based on the implementation in the `/slide` pad. A new button is added under the "*More actions*" drop-down menu.

Couple points to note:

- Print button now has a more generic title text "*Print your document or export it as a PDF file*". This would now apply to the presentation pad as well, so translations will need to be updated. Alternatively, a custom button could be made for each type of pad with unique text.

- The final margins and page orientation depend heavily on local printer setup, so have been left to default for now (tested in Firefox, Chrome).

Let me know if any additions/fixes are required.

### DIFF
--- a/customize.dist/src/less2/include/app-print.less
+++ b/customize.dist/src/less2/include/app-print.less
@@ -46,6 +46,38 @@
                     display: block;
                 }
             }
+            // Code app
+            body.cp-app-code {
+                display: block;
+                * {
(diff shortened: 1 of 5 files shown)


ansuz:
Thanks!

PR #305, opened 2018-12-15
Update documentation for Docker usage

### LINKED ISSUE #301: Cryptpad user information is magic hidden state
I'm trying to migrate a Cryptpad installation from one host to another. Cryptpad is running in a container, with a `docker-compose.yml` section that looks something like this:

```
cryptpad:
    image: xwiki/cryptpad
    volumes:
      - ./cryptpad/files:/cryptpad/datastore
      - ./cryptpad/customize:/cryptpad/customize
      - ./cryptpad/pins:/cryptpad/pins
      - ./cryptpad/blob:/cryptpad/blob
      - ./cryptpad/blobstage:/cryptpad/blobstage
      - ./cryptpad/tasks:/cryptpad/tasks
    environment:
      - USE_SSL=false
      - STORAGE=./storage/file
      - LOG_TO_STDOUT=true
    restart: always
```

When I copy the Docker image (with `docker save` and `docker load` to make sure I have the right one) and the contents of all those mounted directories from one host to another, and start up Cryptpad on the new host, all the pads and their data migrate over, but the user accounts don't. When I go to log in with the username and password that worked on the old install, I get told the account doesn't exist. When I re-register with the same user name and password, I get a fresh, empty CryptDrive, and a new file under `pins` that I think corresponds to it.

Is there anywhere that Cryptpad stores state besides those folders I mounted in the container? Are user account identities somehow dependent on the hostname of the server? (Because I am moving to a different hostname.)

interfect:
This should fix #301. I'm putting the lessons I learned after deleting all my users on my instance into the documentation, so it shouldn't happen again.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,3 +21,8 @@ services:
     volumes:
       - ./data/files:/cryptpad/datastore:rw
       - ./data/customize:/cryptpad/customize:rw
+      - ./data/pins:/cryptpad/pins:rw
+      - ./data/blob:/cryptpad/blob:rw
+      - ./data/blobstage:/cryptpad/blobstage:rw
+      - ./data/tasks:/cryptpad/tasks:rw
+      - ./data/block:/cryptpad/block:rw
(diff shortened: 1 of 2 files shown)


ansuz:
Thanks!

PR #309, opened 2019-01-07
Made CSP syntax in nginx example conf valid

promasu:
Made CSP syntax valid. There were arguments for e.g. blob: oder data: missing

### DIFF
--- a/docs/example.nginx.conf
+++ b/docs/example.nginx.conf
@@ -34,15 +34,15 @@ server {
     # Will not set any header if it is emptystring
     add_header Cache-Control $cacheControl;
 
-    set $styleSrc   "'unsafe-inline' 'self' your-main-domain.com";
-    set $scriptSrc  "'self' your-main-domain.com";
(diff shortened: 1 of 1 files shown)


docs/example.nginx.conf
```diff
@@ -34,15 +34,15 @@ server {
...
-    set $mediaSrc "* blob:";
-    set $childSrc "https://your-main-domain.com";
-    set $workerSrc "https://your-main-domain.com";
+    set $styleSrc	"'unsafe-inline' 'self' your-main-domain.com";
+	set $scriptSrc	"'self' your-main-domain.com";
+	set $connectSrc	"'self' https://your-main-domain.com wss://your-main-domain.com your-main-domain.com https://api.your-main-domain.com blob: your-main-domain.com";
+	set $fontSrc	"'self' data: your-main-domain.com";
+	set $imgSrc	    "'self' data: * blob: your-main-domain.com";
+	set $frameSrc	"'self' your-sandbox-domain.com blob: your-sandbox-domain.com";
+	set $mediaSrc	"'self' data: * blob: your-main-domain.com";
```
ansuz: the addition of  `data:` doesn't seem strictly necessary, but I don't think it opens up any risk of attack.

ansuz:
thanks!

PR #312, opened 2019-01-11
Multi arch docker file

mcrosson:
Update the existing docker files to work on x86-64, armv7 (hf), armv8 as a single Dockerfile.

Also includes the following
- Use node 6 stretch as it's the only node 6 version that's slim and available on all platforms
- Pull latest static release of tini from their release page instead of using the package manager (not widely available on all architectures presently)
- Update docker file volumes so all of the volumes from the docs are listed
- Update order of various docker commands to improve caching and re-build performance


ansuz:
Thanks!

Issue #339, opened 2019-03-14
Browser stalls on initialization of new pad - 2.18.0


nicolass67:
Hello, 

I'm encouring a little problem with cryptpad (CP) 2.18.0. My setup worked with a 2.15 version.
The problem appears with a "bare-metal" install or with the docker image. 

The home page of CP  displays , and I can access to the FAQ, features or other html pages. 
But when I click to create a pad (whatever the type) , the initialization page appears but stays so forever. 

I use an apache reverse-proxy in front of CP. Here is the config (stripped the non relevant infos) : 
... 
```
# https://github.com/xwiki-labs/cryptpad/wiki/Apache
ProxyPreserveHost On
ProxyRequests Off
ProxyPass / http://localhost:3000/
ProxyPassReverse / http://localhost:3000/
RewriteEngine On
RewriteCond %{HTTP:UPGRADE} ^WebSocket$ [NC]
RewriteCond %{HTTP:CONNECTION} Upgrade$ [NC]
RewriteRule .* ws://127.0.0.1:3000%{REQUEST_URI} [P]
```

Cache is disabled , apache proxy_wstunnel is enabled and working 
This setup worked for 2.15.0 of CP. 

The logs in debug dont show errors or whatever. Maybe a clue is this : 
` mod_proxy_fcgi.c(916): [client XX.XX.XX.XX:60772] AH01077: declining URL http://localhost:3000/bower_components/ckeditor/plugins/tableselection/styles/tableselection.css, referer: https://XXXXX/pad/ckeditor-inner.html?ver=2.18.0-1552552142642`

I also tried this workaround, without success : https://github.com/xwiki-labs/cryptpad/issues/216 

Any idea of what can have changed and explain this ?

ansuz:
Most of CryptPad's errors are found in the browser console, not the server's logs. Can you check for errors there?

nicolass67:
Hi ansuz, 

From what I see in the Chrome developper tools the last lines are : 

icons.png?t=H8DC | 200 | png | Other | 9.3 KB | 28 ms
-- | -- | -- | -- | -- | --
tableselection.css | 200 | stylesheet | ckeditor-inner.html?ver=2.18.0-1552552142642

It stalls after ... and there is nothing else that 200 http code ...

ansuz:
is that the network tab or the console tab? The console should show errors if something went wrong.

nicolass67:
This was on the network tab , attached the console tab errors 

[...]

PR #344, opened 2019-03-19
Fix preparation of customize folder

### LINKED ISSUE #336: Updating cryptpad
I run a cryptpad 2.16.0 and would like to udpate to the latest cryptpad. This is what I did in the `cryptpad` directory, which is a clone of `github.com/xwiki-labs/cryptpad`:

```
git pull
docker build -t xwiki/cryptpad .
docker-compose up
```

Everything seems to run correctly, but if I connect to the server, I only get a white screen. If I do

```
git checkout 2.16.0
docker build -t xwiki/cryptpad
docker-compose up
```

it works again as before. But also checking out `2.17.0` doesn't work. I searched around for some instructions how to correctly update, but couldn't find them. Any hints?

ineiti:
Following #336, and perhaps #339, this patch doesn't copy the whole `customize` directory to the docker container, but only the `config.js` file. This should allow for easier update of cryptpad.

Closes #336

### DIFF
--- a/container-start.sh
+++ b/container-start.sh
@@ -3,7 +3,7 @@
 # Creating customize folder
 mkdir -p customize
 [ -z "$(ls -A customize)" ] && echo "Creating customize folder" \
-  && cp -R customize.dist/* customize/ \
+  && mkdir customize/ \
   && cp config.example.js customize/config.js
 
 # Linking config.js


ansuz:
I seems like the `mkdir -p` command above will create the customize folder if it doesn't already exist, so the `mkdir` that you changed will always return with an error. This means the lines after will never run.

I think it works for you because you already have a customize folder, but it might break the regular behaviour for new docker images.

I'm breaking up the first command into a few lines with checks, which I think will be more resilient.

ansuz:
https://github.com/xwiki-labs/cryptpad/commit/4ff7683f2342c19d39f8e1d212f62c01e2905e58

ineiti:
Cool - thanks. In fact I didn't even try it :(

PR #347, opened 2019-03-20
optimize Dockerfile

tbe:
The current Dockerfile had multiple issues:

- uses a big debian image
- installation and removal of build dependencies happens in two steps
 - this does not shrink the image size, as all layers are preserved in the final image

Due issues with alpine, i switched to stretch-slim

Also i switched to multistage builds to make the resulting image more cache friendly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,26 @@
-# 6-stretch is the ONLY node 6 release supported by arm32v7, arm64v8 and x86-64 docker hub labels
-FROM node:6-stretch
+# We use multi stage builds
+FROM node:6-stretch-slim AS build
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -yq git jq python
+RUN npm install -g bower
+
+# install tini in this stage to avoid the need of jq and python
+# in the final image
+ADD docker-install-tini.sh /usr/local/bin/docker-install-tini.sh
+RUN /usr/local/bin/docker-install-tini.sh
+
+COPY . /cryptpad
+WORKDIR /cryptpad
+
+RUN npm install --production \
+    && npm install -g bower \
+    && bower install --allow-root
+
+FROM node:6-stretch-slim
 
 # You want USE_SSL=true if not putting cryptpad behind a proxy
 ENV USE_SSL=false
-ENV STORAGE=\'./storage/file\'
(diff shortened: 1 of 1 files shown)


mcrosson:
Builds good on Docker installs with ```arch``` set to ```arm32``` and ```amd64```. Broken on ```arch=arm64```. I did *not* test run as the builds broke on ```arm64```.

The following were found on arm64

- ```Error: could not get uid/gid``` ; needs ```npm config set unsafe-perm true``` to work around the uid-number bug that tends to pop up every couple months on arm64 in particular.
- ```arm64v8/node:6-alpine``` can not be pulled at present. The ```arm32v6``` and ```arm64v8``` Node Alpine images are wildly inconsistent for pulls from the registry, even if they can be built by hand.
- Can this be updated to use ```6-stretch-slim``` with multi-stage builds instead? That will provide a more stable pull for builds on arm platforms while keeping the resulting image much smaller than the standard ```6-stretch``` base.



For reference, this is the ```arm64v8/node:6-alpine``` Dockerfile I used to verify ```6-alpine``` can be built by hand despite not being available as a pull from Docker Hub. https://raw.githubusercontent.com/nodejs/docker-node/6833ee622c5a1d4c054ffaf96115f2a20714f5f3/6/alpine/Dockerfile . Building ```6-alpine``` by hand on ```arm64v8``` allowed the standard cryptpad builds to succeed but I did not do further tests as the base attempt at building failed due to Docker Hub throwing manifest errors.

tbe:
@mcrosson stretch-slim would be an option. I found, that the image for alpine is there, but the tag for the short version is missing, so i opened a bug.

Just for interest, do you have any reference for the mentioned uid number bug?

tbe:
Rewritten the PR to use stretch-slim as base image.

ansuz:
@mcrosson any update on this? We're trying to clear outstanding PRs but we don't have any ARM systems with which we can test either.

mcrosson:
> @mcrosson any update on this? We're trying to clear outstanding PRs but we don't have any ARM systems with which we can test either.

Builds on arm32 and arm64. I have a test deployment on an arm64 vps up at https://pad.lollipopcloud.solutions/ and I don't see anything obviously broken after applying this PR.

ansuz:
Great! Thanks!

PR #350, opened 2019-03-28
Working CSP config for a production site

chmac:
I spent a good hour fighting with the CSP config. I finally got it working by checking the headers on cryptpad.fr and copying them. However, this config does not work on localhost:3000 in development. I figure that having a second example config file might be an easier solution than fixing the headers to work in both cases. I thus submit this PR for your consideration.

Thanks for the awesome work on cryptpad! ❤️

### DIFF
--- a/config.productionExample.js
+++ b/config.productionExample.js
@@ -1,0 +1,362 @@
+/*@flow*/
+/*
+    globals module
+*/
+var realDomain = "some.domain.tld";
+var realHost = "http://" + realDomain;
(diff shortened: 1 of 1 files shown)


ansuz:
I'm glad you spent the time looking into the issue but I'm hesitant to pull in another file which we'll have to maintain. [The wiki](https://github.com/xwiki-labs/cryptpad/wiki) might be a better place to note the differences between local and production installations so as to minimize code duplication.

PR #351, opened 2019-03-28
Avoid overwriting of existing config.js

ineiti:
When fixing the overly-broad copying in https://github.com/xwiki-labs/cryptpad/commit/4ff7683f2342c19d39f8e1d212f62c01e2905e58 , the fix was to check for links (`-h`). Unfortunately I didn't fire up `man test`, so I didn't catch this one when it was proposed.


ineiti:
Already fixed in staging.

PR #353, opened 2019-04-03
Small fixes

ineiti:
Three fixes for the current master branch which are starting to make their way into 'staging' but don't seem to be there yet.

- Fixes the contentSecurityPolicy bug which made spreadsheet fail.
- Exports ports in docker.

Problem is: the `config.js` of the current master-branch is invalid. So if somebody already started a cryptpad server, he'll have an invalid `config.js` :(


ansuz:
Can you make this PR against the staging branch? We do our development there and it makes conflicts less likely

ineiti:
Done - do you have any idea how to cope with the fact of invalid `config.js` in users' repos?

Also, staging seems to break quite a lot of things in the UI - shall I open issues also for staging?

ansuz:
These were good fixes but we addressed the same issues in a different manner at the same time. Thanks for digging into the issue.

PR #373, opened 2019-05-14
Nix packaging

zimbatm:
Hi,

I am part of the NixOS Foundation. As part of the NLNet Zero grant we are helping accepted projects to package their project into nixpkgs. The goal is to provide a unified and composable way for all the grant's project to be used.

This PR includes a few changes that would make it easier to package cryptpad. A WIP package is available over there: https://github.com/NixOS/nixpkgs/pull/61513

The biggest outstanding difficulty that hasn't been addressed is that the project assumes that it's running from withing the source root. While a package generally has a binary installed in the PATH, the source somewhere in the /usr/lib and the data somewhere under /var like /var/lib/cryptpad. If you want cryptpad to be available in other package managers I recommend making these changes. If you want more details on how to apply those let me know.

Cheers,
z

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -45,7 +45,7 @@
     "localforage": "^1.5.2",
     "html2canvas": "^0.4.1",
     "croppie": "^2.5.0",
-    "sortablejs": "#^1.6.0",
+    "sortablejs": "^1.6.0",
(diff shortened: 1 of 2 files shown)


ansuz:
Looks great! Thanks!

PR #429, opened 2019-08-20
Responsive context menu

ClemDee:
Sub menus in drive context menu can overflow the window on Y axis if parent menu is position too low.
This fix ensure that the sub menus will always be visible.


ClemDee:
Didn't realized I was PR to master, staging would be more appropriate

PR #448, opened 2019-09-23
Bump NodeJS version to 10 to fix #447

### LINKED ISSUE #447: Update base image of Docker
The base image of the Cryptpad Dockerfile is  NodeJS which is unsupported since April 2019. This should be updated to something newer

promasu:
Signed-off-by: Adrian Nöthlich <git@promasu.tech>


ansuz:
Thanks!

PR #454, opened 2019-10-09
Kanban: Move “Add Column” to Bottom

lkiesow:
This patch moves the button for adding new columns to kanban boards from
the top-right to the bottom-left. The reasoning behind this is that it's
not very often used (often only at the beginning when setting up the
board) but it is constantly taking space away from the columns. This is
annoying especially on mobile devices where it even moves all of the
columns to the left, leaving a larger border on the right.


yflory:
I've manually merged your pull request into our "staging" branch.
This will be included in the next release.

PR #455, opened 2019-10-09
Kanban: Move “Add Item” to Header

lkiesow:
This patch moves the button to add new items to kanban boards to the
column headers.

The reasoning behind this change is that on larger boards the buttons at
the bottom may be pushed far away from the actual content, causing users
to constantly scroll up and down over the whole page just to add new
items. See the real-world example below.

This also helps on mobile devices since you may not see all of the
columns there due to the limited screen space. This means that you have
to remember which column you are working on to not accidentally add
items to the wrong column. This patch makes it clear to which  column
users add items since the button now resides next to the column title.

---

Example illustrating the issue to solve:

Screenshot from 2019-10-09 22-05-31

### DIFF
--- a/www/kanban/jkanban.js
+++ b/www/kanban/jkanban.js
@@ -395,17 +395,16 @@
                             contentBoard.appendChild(nodeItem);
                         }
                         //footer board
-                        var footerBoard = document.createElement('footer');
                         //add button
                         var addBoardItem = document.createElement('button');
-                        $(addBoardItem).addClass("kanban-additem btn btn-default fa fa-plus");
-                        footerBoard.appendChild(addBoardItem);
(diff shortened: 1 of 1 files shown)


lkiesow:
Hi, any comment about this?

yflory:
Hi, sorry for the late answer, there was another issue with kanban we wanted to fix first.

I've manually merged your pull request into our "staging" branch. I had to remove the hard-coded "Add new item" string but everything else looked good.
This should be included in the next release.

lkiesow:
@yflory, for future pull-requests, shall I point them towards `staging` instead of `master`?

yflory:
Yes please, `staging` is where we push all our unreleased code and it's easier for us to review and test it.
If it's a PR about documentation or readme, `master` is fine.

PR #465, opened 2019-11-20
fix #460 Cryptpad view mode shows nothing in portrait mode on mobile …

### LINKED ISSUE #460: Cryptpad view mode shows nothing in portrait mode on mobile phones
I only see a blank page when I visit the following link from my phone (Pixel 3a running Android 10) using either Firefox (68.2.0) or Chrome (10.81.10.21.arm64):

  https://cryptpad.fr/code/#/2/code/view/nBYQ0gKZrILo1A5+yLEwpcHON1CoM-+CMxChRPZH34k/embed/present/

If I switch to landscape mode, then I see the expected content.

Perhaps this has something to do with the huge margins.

I was hoping to use a cryptpad to share relevant information with a small group of not technically savy people.  Unfortunately, this is a deal breaker for them.

wolfskaempf:
…phones

Side effects: 

* It is now possible to show the markdown preview on very small screens. It might be shown by default.
* On very large screens the text in the preview will become harder to read as one line spans across the entire screen

### DIFF
--- a/www/code/app-code.less
+++ b/www/code/app-code.less
@@ -142,7 +142,10 @@
             flex: 1;
             max-width: 100%;
             resize: none;
-        }
+         }
+         #cp-app-code-preview-content {
+             max-width: 100%;
+         }
         #cp-app-code-preview {
             // display: none !important;
         }
@@ -155,4 +158,3 @@
         .markdown_gfm-table(black);
     }
 }
-
--- a/www/code/app-code.less
+++ b/www/code/app-code.less
@@ -120,7 +120,7 @@
     }
 
     #cp-app-code-preview-content {
-        max-width: 40vw;
+        max-width: 100%;
         margin: 1em auto;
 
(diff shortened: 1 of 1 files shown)


www/code/app-code.less
```diff
@@ -144,7 +144,7 @@
             resize: none;
         }
```
yflory: You can add the `max-width` rule here and keep `40vw` above to avoid the second side-effect.
```suggestion
         }
         #cp-app-code-preview-content {
             max-width: 100%;
         }
```

wolfskaempf:
Don't merge this yet, it still has some side effects in its current state.

ansuz:
I've modified this PR to be based against our staging branch where we do our development.

ansuz:
I've closed this PR since it's been open without any modifications for a long time. Feel free to re-open if there are any updates.

PR #483, opened 2020-01-03
Removed duplicate location ^~ /datastore/ block

MacLemon:
The ```location ^~ /datastore/``` block was duplicate, meaning nginx would have balked on it during config check.


MacLemon:
Already obsoleted in `staging` branch.

PR #484, opened 2020-01-04
Log errors in faulty custom config file

promasu:
CryptPAD now terminates and does not fallback to default config if the custom config is faulty.
The error in the custom config is logged to the terminal now.

Signed-off-by: Adrian Nöthlich <git@promasu.tech>


ansuz:
Thanks!

PR #503, opened 2020-03-10
Check httpSafeOrigin in config variable

101100:
Noticed I was getting the warning and tracked it down to a missing `config.`


ansuz:
Good catch! Thanks!

PR #513, opened 2020-03-21
Add shutdown() to non-disk logger

7adietri:
The script `evict-inactive.js` calls `Log.shutdown()` at the end, which currently fails if logging to disk has been disabled via `logPath`.


7adietri:
Disclaimer: I am not a NodeJS developer.

ansuz:
This seems safe. Thanks!

PR #520, opened 2020-03-30
Update nginx-configuration

kitoy30:
Tested on Debian buster; Nodejs v12.16.1: 

The properties 'unsafe-eval' and 'unsafe-inline' in script-src causes bugs with CSP, preventing Sheets to work when installing Cryptpad on self hosting.
Commenting them is enough to work most cases, I let them commented in case someone wants to test other configurations.


ansuz:
Those rules are there for a reason. Our staging branch has a fix for the root cause as explained in [this issue](https://github.com/xwiki-labs/cryptpad/issues/518).

PR #525, opened 2020-04-11
Allow config.defaultStorageLimit to be zero bytes

ma-neumann:
Hi cryptpad-team,

since commit f86196e40adfccade922e63bba5dafbff20538fe config.defaultStorageLimit cannot be zero anymore. I suppose this might break some configurations.

While https://github.com/xwiki-labs/cryptpad/blob/c8c98b57479637484b36ba43e14c45671cdcffe8/lib/load-config.js#L35 allows it to be >= 0, https://github.com/xwiki-labs/cryptpad/blob/f86196e40adfccade922e63bba5dafbff20538fe/lib/historyKeeper.js#L82 requires it to be > 0.

Greetings,
Alex

### DIFF
--- a/lib/historyKeeper.js
+++ b/lib/historyKeeper.js
@@ -98,7 +98,7 @@ module.exports.create = function (config, cb) {
     paths.staging = keyOrDefaultString('blobStagingPath', './blobstage');
     paths.blob = keyOrDefaultString('blobPath', './blob');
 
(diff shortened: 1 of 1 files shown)


ansuz:
Good catch! In the absence of code to restrict access without registration I don't think a storage limit of 0 will do what people expect. We plan to introduce invite-only registration and the ability to restrict usage of a server in the coming months, so we'll definitely revisit this set of configuration values. In the meantime there's no harm in making those checks match, so this seems ok to merge.

PR #537, opened 2020-04-26
Add privacy policy option

7adietri:
Adds a `privacy` option to `application_config_internal.js`, similar to `imprint`.


7adietri:
BTW, customizing the imprint via `customize/imprint.html` doesn't seem to work. It redirects to `/imprint.html/` and returns a 404.

PR #538, opened 2020-04-26
List premium features only with subscriptions

7adietri:
Shows the premium features panel on `/features.html` only when subscriptions are allowed.


yflory:
I've merged the PR manually into our `staging` branch, it will be included in the next release.
Please note that subscriptions are now only enabled on cryptpad.fr

7adietri:
> Please note that subscriptions are now only enabled on cryptpad.fr

Yeah, that's why I thought it might be less confusing to hide the list on other instances. Thanks!

Issue #609, opened 2020-09-07
messages.js - referring to Finnish as a language


ilogreensward:
Hello,

I would like to report a minor translation issue in customize.dist/messages.js pertaining to the language selector. 
The Finnish language is referred incorrectly as "**Suomalainen**" (adjective), while the correct proper noun for the language is **Suomi**.

Thank you and have a good day.

sisco0:
Have a good day, the pull request has been submitted in order to overcome this issue.
I do not know very much about Suomi, but I must admit that _Nightwish_ was a great music band. 😄

ilogreensward:
Thank you for fixing this one!

PR #615, opened 2020-09-16
Finnish language renamed in a correct way

sisco0:
Have a good day.
As cited at #609 there is a problem at naming _finnish_ so this change is needed in order to provide correctness.


ansuz:
Thanks!

PR #618, opened 2020-09-22
Unsafe Origin for CSP modified and Migration script fixed

sisco0:
In order to be able to load the webpage locally using `localhost` or `127.0.0.1` this change is required.
Content Security Policy could accept different domain names using whitespace as separator.
When running `npm start` at the shell the server is loaded at `127.0.0.1` but localhost was the one specified at the _Unsafe Origin_ configuration example.
In this way, both `localhost` and `127.0.0.1` could be used to access the current development server.

### DIFF
--- a/config/config.example.js
+++ b/config/config.example.js
@@ -46,7 +46,7 @@ module.exports = {
  *  cryptpad/docs/example.nginx.conf (see the $main_domain variable)
  *
  */
-    httpUnsafeOrigin: 'http://127.0.0.1:3000/',
+    httpUnsafeOrigin: 'http://127.0.0.1:3000/ http://localhost:3000',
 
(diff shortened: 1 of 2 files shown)


sisco0:
Migration script has been fixed as well, there was the need of adding the `err` parameter at a function.

ansuz:
Sorry for taking so long to review this. It's been a very hectic year and I'm just now catching up on PRs.

I've merged your patches into our `staging` branch with some modifications.

1. The fix for the migration script correctly accounts for a change in an API, but I converted its use of tabs to spaces to match the rest of the codebase.
2. I decided to preserve the existing default in the example config to avoid confusing admins, but added your multiple origin example in the comment above so anyone interested in the same behaviour has an example.

Your patches and my amendments will be included in our next release.

PR #630, opened 2020-11-05
Set reasonable env value for $PWD

pietsch:
/home/cryptpad/cryptpad/cryptpad seems one cryptpad too many, and it does not match the sample value of WorkingDirectory above


ansuz:
Indeed, good catch!

PR #647, opened 2020-12-01
Support non-writable working directory

angdraug:
Mixing code and data in the same writable directory is a security risk. Mitigating it requires running CryptPad as a user that doesn't have write permissions to the source tree, and configuring all data paths to be under a directory that user can write (e.g. source under /home/cryptpad/cryptpad and data under /var/lib/cryptpad).

Without these patches, some places in CryptPad fall back to PWD based paths and fail with EACCESS.

### DIFF
--- a/config/config.example.js
+++ b/config/config.example.js
@@ -295,6 +295,8 @@ module.exports = {
      */
     blobStagingPath: './data/blobstage',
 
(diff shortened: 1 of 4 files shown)


ansuz:
Thanks! I've merged this into our `staging` branch, which will be included in our next release.

PR #667, opened 2020-12-20
Fixed link

jwehmschulte:
Changed a link from an empty wiki-page to one I thought should be there. I'm assuming the contribute page changed at some point and someone forgot to fix the link?


ansuz:
Thanks!

PR #718, opened 2021-04-20
Add option to open links on first click.

nilsgabriel:
Adds an option to settings panel where the user can decide if they want to open links directly in a new window within the rich text editor.


ansuz:
Thanks! Overall the code looks good. There are some minor things that we'll adjust before deploying in our next release. Certain versions of Chrome will crash if you try to open a link from inside an iframe when certain HTTP headers are set. You might notice that we handle the click by opening our "bounce" app which handles this and other edge cases.

nilsgabriel:
Oh, that was fast. Thank you. I'll have a look the „bounce“ app.

ansuz:
On further testing it looks like [ths line](https://github.com/xwiki-labs/cryptpad/commit/df35c21a1a0eba8a8d4c45c24ec94d8206981340) references framework which is not guaranteed to exist at the time when this is called. Other functions in the same scope reference it, but only for the purpose of registering handlers which will be called by the time framework is actually defined.

nilsgabriel:
Hmmm... Is there another way to access saved settings? I was searching for an „easier“ option to get the current settings object, but coudn't find one.

PR #741, opened 2021-06-05
Improve Markdown Heading Styling

zicklag:
Hey there! This project is _very_ cool!

I just found this today and _might_ be contributing more if I can find the time. For now, here's a quick PR to ( subjectively ) improve the heading sytle in the markdown editor.

### Before

image

### After

image

### DIFF
--- a/customize.dist/src/less2/include/markdown.less
+++ b/customize.dist/src/less2/include/markdown.less
@@ -163,6 +163,10 @@
         font-weight: bold;
         padding-bottom: 0.3em;
         border-bottom: 1px solid @cp_markdown-border;
+        margin-top: 1em;
+    }
+    h4, h5, h6 {
+        border-bottom: none;
     }
     li {
         min-height: 22px;


zicklag:
I also just pushed an update to the CkEditor heading styles which were previously very hard for me to read the different heading levels because they looked the same as unstyled text.

### Before

image


### After

image

zicklag:
For some reason the CkEditor styles aren't applied when printing. I'm trying to figure that out now.

zicklag:
Just pushed a fix for the printing styles not being picked up, and I tweaked the ckeditor styles just slightly more to make it match the markdown preview styles:

image

---

Also, just curious, if this gets merged, what is the normal release cadence, or when I could expect this to get published to the live instance at cryptpad.fr?

ansuz:
I quickly reviewed this PR before our 4.7.0 release and noticed that it seemed to also affect styles in the code editor itself. It wasn't obvious why this would be the case and I haven't had the opportunity to dig into the problem yet, but I intend to so this can get merged.

For future reference, this would be easier to review if it were split into two PRs (for rich text and code), but that's not a blocking issue.

ansuz:
PS: thanks for the effort in the meantime!

zicklag:
> For future reference, this would be easier to review if it were split into two PRs (for rich text and code), but that's not a blocking issue.

Gotcha. I'll do that in the future.

> PS: thanks for the effort in the meantime!

No problem! Thanks for CryptPad! :)

PR #749, opened 2021-06-13
Sso prototype

ClemDee:
This PR results from an end-of-study project, and aims to add a SSO solution for enterprises for the CryptPad software.

This is at prototype stage, but can still be used as a good starting point for a further implementation into the software.

One limitation due to the nature of the Cryptpad software, is that the user still have to provide a dedicated Cryptpad password after the SSO login / registration, as this password is needed to generate their public and private keys. 

## Basic features:
- Restrict page access via nginx, and redirect user to the SSO portal
- Adapt login / register pages to SSO mode
- Restrict user rights (can't rename, can't delete account, ...)
- Automatically display other registered LDAP users in the "share" panel

## How to configure 

The following steps are required in order to be able to setup and test the PR changes.

First, setup a working normal CryptPad instance. This will be our base to setup the SSO.

In order to support SSO inside Nginx, we used the [lua-resty-openidc](https://github.com/zmartzone/lua-resty-openidc) library.
However, in order to have lua scripts to work inside Nginx, we need to use the [OpenResty](https://openresty.org/en/) enhanced version of Nginx:

```sh
$ sudo apt-get -y install openresty openresty-opm
```
Then download `lua-resty-openidc` using the Openresty Packet Manager:
```sh
$ opm install zmartzone/lua-resty-openidc
```

The [Nginx configuration file](https://github.com/xwiki-labs/cryptpad/blob/main/docs/example.nginx.conf) needs to be replaced by the following: [sso.example.nginx.conf](https://github.com/xwiki-labs/cryptpad/files/6644571/sso.example.nginx.conf.txt)

You will also need the following file, to be placed in `/etc/nginx/`: [oidc.conf](https://github.com/xwiki-labs/cryptpad/files/6644576/oidc.conf.txt)
In this file you will be able to configure your sso endpoints.

Also, in the `customize` folder of your CryptPad instance, override the `application_config.js`, and add the following
```js
AppConfig.ssoEnabled = true;
```
so that SSO is enabled everywhere inside the instance.

Once you're done modifying the configuration files, you can start everything:
```sh
$ service openresty restart
$ node server.js
```

## Further work

For now, only LDAP users of the server can have access to the instance. It could be great to also implement a system of guest user, who could have view rights on given pads.


vncntwbr:
What's the status with respect to LDAP implementation?

PR #767, opened 2021-07-06
Debloated contact info

comradekingu:


ansuz:
The Code of Conduct is there for a reason.

comradekingu:
Got the update about new strings. Wanted to volunteer some time and noticed I had fallen behind.

Maintainers usually don't have any intentions in putting this up
beyond getting green lights for the GitHub checklist or following some guide saying "it is important".
I thought since it was the old 1.4 version that was the case here.
The updated one makes for some interesting reading for the select few people that actually do.
Predictably "project maintainers" is replaced by "community leaders".

I must have contributed without checking and
never have nor will pledge allegiance to this. It would have had to be because its importance is
that of easily being overlooked, or otherwise tied to some function as an EULA. Will remember not to contribute now though.
The testable case is to allow people to voluntarily pledge allegiance to it, or even ask in the first place.
Out of the 500+ projects I have contributed to, that hasn't happened once.

Having gone through the trouble of changing the master branch naming scheme to "main",
I don't think these helping features/hurdles to the ecosystem on the premise of making something
better can do any better than demonstrably changing things for the better, or worse.
I am otherwise always listening to arguments for doing any of it, but those are hard to come by.

The Covenant creators are asking for monetary contributions if you use their code of conduct.
Somehow people have as little faith in this initiative as the others. The PR is there for the taking in every respect.

ansuz:
> Will remember not to contribute now though.

Thanks!

PR #853, opened 2022-01-04
Example NGINX config: allow embedding only from main_domain

innir:
While `frame-src` already restricts what `main_domain` can embed the other direction was missing. Adding `frame-ancestors` closes the loop - only content from `sandbox_domain` can solely be embedded from `main_domain`.

The second commit implements the same measure for old browsers - I just added it for completeness but I can remove it if you think it is superfluous.

Tested for some days now with no regression (I also tested the new OnlyOffice components).


innir:
This is basically implemented with https://github.com/xwiki-labs/cryptpad/commit/e1abf4ef7722f8eca9214b4db770b945213ff2a1 so I'll close this pull request.

PR #891, opened 2022-03-28
Attempting to integrate codemirror wrap/hardwrap addon.

JLiu1272:


www/settings/inner.js
```diff
@@ -1569,6 +1570,62 @@ define([
...
+            if (!isNaN(val) && typeof (val) !== 'number') { return; }
+            common.setAttribute(['codemirror', key], val);
+        }).appendTo($container);
+
+        $cbox.find('input').change(
+            function () {
+                // If enable hard wrap max width is enabled, then disabled=false
+                // otherwise, disabled=true 
+                $input.attr('disabled', !this.checked);
+                common.setAttribute(['codemirror', hardWrapMaxWidthEnabledKey], this.checked);
```
JLiu1272, line 1608: Made it so that when this feature is disabled, the input to change the width for wrapping is also disabled, and vice versa.

ansuz:
Hi,

I finally got around to reviewing and testing this in more depth. I made [some changes](https://github.com/xwiki-labs/cryptpad/commit/7ce81912324bb3912c021ae3ede10c689f7467e2) in the process. I added some comments (via GitHub's UI, not in the code) which explain my reasoning.

This PR does exactly what it's supposed to do when editing alone, however, I found that it introduces some problematic behaviour in collaborative editing sessions. The `on('change', ...` handler is triggered when a change is made to the editor, but that doesn't distinguish between changes you made and those which were made by remote editors.

The consequence of this is that if one person joins an editing session with hard-wrapping enabled their client will wrap lines according to their preferences any time anybody makes a change, which could be very disruptive.

Another way to approach this is to allow users to set their preferred wrapping width in settings and to apply that to documents that they create, but to provide an additional setting for individual documents that 

1. enables this to be easily enabled/disabled. 
2. ensures that clients agree on the formatting of the shared document

We do this to some degree for the _author colors_ functionality.

JLiu1272:
Hi, sorry, got wrapped up with other things. Thanks for the feedback!

PR #925, opened 2022-06-24
update example.nginx.conf to listen on ipv6 too

mrbesen:
A admin only using ipv4 might overlook this configuration entry and make his cryptpad-instance only available on ipv4. Unintentionally locking-out every ipv6 user.

This change allows ipv4 and ipv6 access.


mthld:
Looks good to me, what do you think @ansuz? Maybe it needs to be merged in another branch tho.

PR #931, opened 2022-07-19
Opengraph

Bad89:
thanks need to learn


ansuz:
hi,

can you explain what this is intended to do?

PR #938, opened 2022-08-03
Issue templates

mthld:
Here is a work-in-progress to update our issue templates. As we discussed internally I'm sticking to the new workflow we were thinking about.

**Basically**
- An issue template for bug resolution, `bug_resolution.md`
- An issue template for a new feature, `new_feature.md`

**Everything else is links to the forum** (see `config.yml`)
- Installation, configuration issues
- Bug reports
- Feature request

```
⚠️ we'll need to update the links to match the proper forum category!
```

Also, if we generate a `SECURITY.md` file, there will automatically be a 6th entry linking to this file.

Please don't hesitate to amend &/or modify these templates!


mthld:
As discussed with Aaron last week, since everything forum related wouldn't be done soon, I modified the templates to leverage the current organization.

PR #967, opened 2022-09-17
Add reminder to remove adminKeys block comment

jmlutgemeier:
Hey, I just created a reminder to remove the block comment as its very easy to overlook it which creates headache!


mthld:
Hello!

Thank you very much for your proposal and your time invested in finding a solution to this.

After discussions with the team we won't be accepting this PR. In place we will likely remove the block comment, while adding an example above.

Thank you again, I'll be closing this now! :)

PR #972, opened 2022-10-07
Make unparseable message non-fatal

dbkr:
There was code to handle unparseable messages from things like third party browser extensions but it caught a JSON parsing error, logged an error and then continued, which would throw an unhandled exception trying to read the `ack` property of `data` which had been left as `undefined`.

Add a `return` to abort execution of the handler function and ignore the message instead. Also changes the error to warning, since it's not a fatal error. Add a comment to clarify the behaviour.

From the existing comment, it looks like this was the original intention of the code, but the `return` was simply missed.

### DIFF
--- a/www/common/outer/worker-channel.js
+++ b/www/common/outer/worker-channel.js
@@ -171,10 +171,12 @@ define([
             var data;
             // apparently some browser extensions send messages to random targets
             // which can trigger parse errors that interrupt normal behaviour
+            // we therefore log a warning and ignore any messages we can't parse
             try {
                 data = typeof(msg.data) === "object" ? msg.data : JSON.parse(msg.data);
             } catch (err) {
-                console.error(err);
+                console.warn(err);
+                return;
             }
             if (typeof(data.ack) !== "undefined") {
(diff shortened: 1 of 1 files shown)


mthld:
Hello,

Many thanks for your contribution!
We'd like to know how did you received such message? Can you describe what happened?

Thanks!

dbkr:
It was caused by a third party browser extension ([Superagent](https://www.super-agent.com/), in this case) which uses postmessage to communicate between its various components. One of these is a message to the page with just the string `GET_URL`, which of course doesn't parse as JSON.

The way this manifested for me was when opening a rich text document, a full-screen modal message saying that GET_URL isn't valid JSON would appear. This is then replaced by a similar message with an error about reading the 'ack' of `undefined` (ie. when evaluating `data.ack` below). This app would then remain on this error screen so I couldn't view or edit the document.

ansuz (approved):
Good catch! You're correct that this was the intended behaviour, but I must have been in a bit of a rush. It looks good to me, and I like that you included a comment explaining it :)

Issue #295, opened 2018-11-20
bower deprecation warning


anadahz:
While building cryptpad I get the following deprecation warning:
```
npm WARN deprecated bower@1.8.4: We don't recommend using Bower for new projects.
Please consider Yarn and Webpack or Parcel.
You can read how to migrate legacy project here: https://bower.io/blog/2017/how-to-migrate-away-from-bower/
```

ansuz:
We will very likely move to one of these other package managers at some point in the future, but it seems like it's a safe tool for now.

zimbatm:
The bower dependency makes it quite difficult to package the project for me. Because bower is deprecated the tooling to generate nix code from the bower.json is not very well maintained. bower2nix is quite slow and the output needs to be tweaked after the fact as well.

Not sure if this is a motivation for you but it would make my life much easier :)

ansuz:
I much prefer npm to bower. I'm using it in combination with browserify for my personal projects.

The main things stopping us is that moving from bower to npm will take a significant amount of time and that for most users it won't improve their experience of using CryptPad. We've also discussed moving away from requirejs and implementing a build system, which _would_ directly contribute to visible improvements, but this is an even bigger job.

We're waiting to hear back about some funding which would directly finance this work, but until we know more I can't promise to fit it into our roadmap.

zimbatm:
Thanks I appreciate that it's not a priority.

For reference https://github.com/NixOS/nixpkgs/pull/61513 is adding cryptpad to nixpkgs. Right now there is too much manual patching to automate the upgrade process so the package will likely be stuck at that version until bower gets removed.

ansuz:
2.21.0 isn't the worst version to get stuck at, I suppose.

If that funding comes through then we should be able to drop bower by Autumn 2019, hopefully we'll be able to resume the NixOS packaging then.

blitz:
@ansuz Do you have any news? I'd love to automate the update process for the NixOS package and the bower dependency still prevents this.

Pamplemousse:
Hi there, bumping this thread again:

Although Bower is still maintained on paper, its developpers recommend not to use it for new project (see https://bower.io/).
It means less and less projects are relying on it, and tooling around it is becoming obsolete, thus unmaintained.
For example, the [nix](https://nixos.org/) ecosystem used to rely on [`bower2nix`](https://github.com/rvl/bower2nix) to package bower components, but it hasn't been updated in the last 4 years, and has several blocking issues by now...

[...]

PR #989, opened 2023-01-07
[WIP] Debowerify

Pamplemousse:
`bower` is deprecated to the point that `cryptpad` is being dropped in places because of its maintenance burden (see #295).

This PR aims at removing the need for `bower` for building `cryptpad`.

- [ ] Naively migrate libraries to use `npm` and custom build script
  - [x] in development environment
  - [ ] run tests
  - [ ] for production build
- [ ] Use a JS bundler instead of `npm` + custom copies: relates to https://github.com/xwiki-labs/cryptpad/issues/881

It's my first time contributing to `cryptpad`, and I am not familiar with the codebase, so I will appreciate any feedback and help for testing :slightly_smiling_face:, thanks!

Relates to:
- xwiki-labs/chainpad-listmap#3
- https://forum.cryptpad.org/d/75-missing-dependency-secure-fabricjs-in-wwwwhiteboardexportjs


Pamplemousse:
I have been trying to make it work in a development environment. Seems that I am mostly there, but there is still something trying to be loaded from `www/bower_components/` in the index page, and I can't figure out why...

```html
<script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="/bower_components/tweetnacl/nacl-fast.min.js" src="/bower_components/tweetnacl/nacl-fast.min.js?ver=5.2.1-1673098798860-1673098953727"></script>
```

yflory:
I was able to "fix" this `bower_components` call from the index page here: https://github.com/xwiki-labs/cryptpad/blob/debowerify/www/common/requireconfig.js#L31 It was coming from the `chainpad-crypto` library.

I've also published `chainpad-listmap` to npm and fix the fabricjs issue on our `main` branch.

Otherwise the PR looks good but still needs further testing on our side

Pamplemousse:
> I was able to "fix" this bower_components call from the index page here: [debowerify/www/common/requireconfig.js#L31](https://github.com/xwiki-labs/cryptpad/blob/debowerify/www/common/requireconfig.js?rgh-link-date=2023-02-01T16%3A13%3A53Z#L31) It was coming from the chainpad-crypto library.

Thanks!
Could it be a single instance of a more recurrent issue though? Every library `cryptpad` uses could have `bower` dependencies that would introduce a similar problem, right?
Is it OK given that the distribution / loading of dependencies should be revamped to use a JS bundler eventually?

PR #990, opened 2023-01-10
Date input in forms

zuzanna-maria:
As requested [here](https://github.com/xwiki-labs/cryptpad/issues/811), this provides a way to include questions in forms where the desired answer format is a date.

### DIFF
--- a/www/form/inner.js
+++ b/www/form/inner.js
@@ -200,7 +200,6 @@ define([
         ];
     };
 
-
     var editDateOptions = function (cb) {
         var evOnSave = Util.mkEvent();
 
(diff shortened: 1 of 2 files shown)


www/form/inner.js
```diff
@@ -1995,6 +2004,75 @@ define([
...
+                    var answer = obj.msg[uid];
+                    if (!answer || !answer.trim()) { return empty++; }
+                    results.push(h('div.cp-charts-row', h('span.cp-value', answer)));
+                });
+                return h('div.cp-form-results-contained', h('div.cp-charts.cp-text-table', results));
+
+
+            return h('div.cp-charts.cp-bar-table', results);
+        },
+        icon: h('i.cp-calendar-active.fa.fa-calendar')},
```
davidbenque, line 2075: this line is causing UI issues
- the class `cp-calendar-active` is only used in the calendar app, not needed here
- the icon is not aligned properly in the UI (see on hover) which could be due to the fact that it's the only icon to use `fa` while others use `cptools` there might be a CSS selector to correct

PR #997, opened 2023-01-19
Hide editor if previewing code read only and it is previewable

henrystern:
This is useful when you just want to share the compiled markdown with someone who would be distracted by the raw text.

I also changed the slider styling, this makes it more obvious that you can drag it to adjust the panes.


henrystern:
Oh cool that's great! I had seen that option but assumed it was for giving present permissions in a presentation or something.

Thanks a lot, I'm enjoying the app.

PR #1003, opened 2023-02-09
Auto-select document name on edit if it's still the default

piemonkey:
A small convenience to enable clicking a document name and immediately typing the name, if it hasn't been set already.

Since the background colour of the `input` is different to the text that it replaces, it always tricks me into thinking it's selected, so I just start typing. The text then just gets appended to the default name.


davidbenque:
Thank you for this!

Issue #752, opened 2021-06-15
Move the Add document button to the left


Th0rgal:
Hello,

Thank you very much for this tool, I love it! I have a suggestion for the drive because I find the position of the "add document" button really annoying. When I add a lot of documents at the same time it is constantly moving. Would it be possible to put it at the start of the list instead of the end?

![image](https://user-images.githubusercontent.com/41830259/122034093-746b9a80-cdd1-11eb-8d89-c458f9997d90.png)

Thanks again,
Thomas

davidbenque:
The screenshot does not include it, but there is a New button in the toolbar which offers the same functionality in almost the same place

<img width="745" alt="218880410-2d1172dc-02a6-4ee9-97e7-955e32c97325" src="https://user-images.githubusercontent.com/11028490/219338737-eecf7e27-3f30-4ef0-9c03-94d41398433a.png">

PR #1006, opened 2023-02-14
This change makes the New button to go on the top of the list. Addresses #752

redchrision:
The New button in CryptDrive will appear in list view on the top of the list instead of the bottom of it, as shown in Image_1. In grid view, this will place the New icon before of the File names list, as shown in Image_2. 

This is addressing Issue #752. 

Image_1
Image_2

### DIFF
--- a/www/common/drive-ui.js
+++ b/www/common/drive-ui.js
@@ -3478,7 +3478,7 @@ define([
             var isInRoot = currentPath[0] === ROOT;
             var $element = $('<li>', {
                 'class': 'cp-app-drive-element-row cp-app-drive-new-ghost'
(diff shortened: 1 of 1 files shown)


davidbenque:
Thanks for the contribution. Unfortunately we will not be merging this as moving this button doubles functionality with the New button in the toolbar (see screenshot). The current position exits to suggest that you can add something at the _end_ of the list or grid of icons, admittedly not very useful but IMO if anything the button should be removed not moved.

The issue that you're responding to is based on an old UI and obscures this part of the toolbar.

218880410-2d1172dc-02a6-4ee9-97e7-955e32c97325

redchrision:
Thank you! It makes sense, indeed.

PR #1018, opened 2023-04-05
[NEW CONTROL] Paragraph + Description

ErivanDev:
A new control (textarea2)  in Forms.

A control that joins the Paragraph and Description.

image


davidbenque:
Thank you for this proposal. I think I see where you are coming from, the need to add additional information to a paragraph question. However it's easily achievable with a description and paragraph blocks, so it doesn't seem worth it to add a question type for this use-case. The editing interface is a bit confusing as well IMO.

PR #1028, opened 2023-05-07
provide visual feedback that a dragged item can be dropped

ansuz:
> provide visual feedback that a dragged item can be dropped into the drive's tree when hovering

I noticed a while back that dragging an element from the drive's main pane to the tree did trigger any visual feedback while hovering with the element. This made it somewhat unclear whether the item was able to be dropped. Aside from the visual cues, everything does seem to work as expected when the element is dropped, so we just need better styles.

This PR adds those styles to the main document tree, the templates category, and the trash - but not on un-droppable categories like "recent" or "tags".

The styles are applied using a less mixin, so it should be simple to add any other styles, such as a contextual cursor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,7 @@ To upgrade:
 
 ```bash
 git fetch origin --tags
-git checkout 5.3
+git checkout 5.3.0
 ```
 
 1. Restart your server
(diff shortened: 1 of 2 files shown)


ansuz:
:heart:

PR #1057, opened 2023-06-13
Fix module not found error

youshitsune:
On some machines when you try to run your server, it throws error module not found.

This is fix to it.

### DIFF
--- a/lib/eviction.js
+++ b/lib/eviction.js
@@ -3,9 +3,9 @@ var Bloom = require("@mcrowe/minibloom");
 var Util = require("../lib/common-util");
 var Pins = require("../lib/pins");
 var Keys = require("./keys");
-var Path = require('node:path');
+var Path = require('path');
 var config = require("./load-config");
-var Fs = require("node:fs");
+var Fs = require("fs");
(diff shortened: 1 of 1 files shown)


youshitsune:
I said on some machines because I didn't test it on other Linux Distros,but on two of mine machines (Debian 12 and Debian 11), I got error module not found in eviction.js file.

Error was for ```require("node:path")``` and ```require("node:fs")```.

ansuz:
The intention behind this style of import is to make it clear that a built-in module is being imported, and to make it difficult to accidentally import a third-party module that has been maliciously [typo-squatted](https://en.wikipedia.org/wiki/Typosquatting).

This style of import is supported in the version of NodeJS which is [recommended in the documentation](https://docs.cryptpad.org/en/admin_guide/installation.html) (16.14.2 or later). In fact, that part of the documentation is due for an update, as version 16 is [nearing the end of its long-term support (LTS) status](https://nodejs.dev/en/about/releases/) and will not have security fixes backported beyond 2023-09-11.

I suggest installing a newer version of NodeJS via NVM as recommended, and continuing to use the `node:fs` syntax.

PR #1065, opened 2023-06-25
add form title to archived notifications

### LINKED ISSUE #1060: Form title is missing on notifications page
### Description

On the notifications page `/notifications/`, the history shows "Unknown document" instead of the form name for notifications about new responses.

### Steps to reproduce

1. Navigate to `/notifications/` and select the History tab
2. Observe that any notifications about new responses to a form display "Unknown document" instead of the title of the form

### Relevant logs

_No response_

### Expected behavior

The title of the form should be displayed in notifications history

### What opreating system are you using?

_No response_

_No response_

### Extensions

uBock Origin, Passbolt

### Version

5.3.0

lemondevxyz:
Fixes #1060 

Notifications in CryptPad have two versions AFAIK, one that sends the notification to the active WebSocket and one that is stored in ./datastore/{HASH}/{HASH}.ndjson

Theoretically, the notifications sent by ./www/form/inner.js are privitized by the server when archived. I haven't confirmed this but it seems that the server, when sending the websocket notification, either fetches the form's data or removes it when storing the notification in data store. (your guess is as good as mine)

Anyhoo, this fix is a bit hacky since it just adds a title field to the notification:
```js
// ./www/form/inner.js:3597
// previous
msgData: { channel: priv.channel }
// current
msgData: { channel: priv.channel, title: framework._.title.title || framework._.title.defaultTitle }
```

Then, it assigns the title in the "right" place because a title in the data field of a notification will not be used.
```js
// ./www/common/sframe-common/mailbox.js:270
// previous
messages.push({
	// ... other fields
	content: {msg: data.message, time: data.time, hash: data.hash},
	// ... other fields
})
// current
messages.push({
	// ... other fields
	content: {
		// ... other fields
		msg: Object.assign(data.message, {
			content: Object.assign(data.message.content, {
				title: data.message.content.data.title,
			})
		},
		// ... other fields
	}
	// ... other fields
})
```

(above code examples are not 100% accurate and are used for presentation purposes...)

### DIFF
--- a/www/common/sframe-common-mailbox.js
+++ b/www/common/sframe-common-mailbox.js
@@ -271,7 +271,11 @@ define([
                     messages.push({
                         type: type,
                         content: {
-                            msg: data.message,
+                            msg: Object.assign(data.message, {
+                                content: Object.assign(data.message.content, {
+                                   title: data.message.content.data.title,
+                                }),
+                            }),
                             time: data.time,
                             hash: data.hash
                         }
--- a/www/form/inner.js
+++ b/www/form/inner.js
@@ -3599,7 +3599,7 @@ define([
                     send: true,
                     anon: true,
                     query: "FORM_RESPONSE",
-                    msgData: { channel: priv.channel }
+                    msgData: { channel: priv.channel, title: framework._.title.title || framework._.title.defaultTitle },
                 }, function (err, obj) {
                     if (err || !obj || !obj.state) { return console.error('ENOTIFY'); }
                 });
@@ -4756,7 +4756,7 @@ define([
             var endDateStr = h('div');
             var $endDate = $(endDateContainer);
             var $endDateStr = $(endDateStr);
-        
+
             var refreshEndDate = function () {
                 $endDate.empty();
 


wginolas (commented):
I tested this and it looks good. @yflory, can you have a look at the code?

yflory:
Hi,
While reviewing your pull request, I found an underlying issue with all notifications from the history which explains the difference between "live notifications" and history. The part of the client that was supposed to add the most recent "title" by searching through the user's drive was not called! I'm currently testing a fix.

> Notifications in CryptPad have two versions AFAIK, one that sends the notification to the active WebSocket and one that is stored in ./datastore/{HASH}/{HASH}.ndjson
> 
> Theoretically, the notifications sent by ./www/form/inner.js are privitized by the server when archived. I haven't confirmed this but it seems that the server, when sending the websocket notification, either fetches the form's data or removes it when storing the notification in data store. (your guess is as good as mine)

FYI, the server has nothing to do with this, it only sees encrypted data. Everything is done in the client!

PR #1067, opened 2023-06-28
add a "make a copy" element to onlyoffice applications

### LINKED ISSUE #1061: Make a copy item not present in menu
### Description

"Make a copy" item isn't present in the CryptPad menu for Spreadsheet, Document & Presentation apps. 

Screenshot 2023-06-21 at 10 14 59

Note that it's still possible to do it by doing a right click on the file from the drive.

### Steps to reproduce

1. Create a Spreadsheet, Document or Presentation file
2. Click the CryptPad File menu on the top left
3. Witness there is no "Make a copy" item

### Relevant logs

_No response_

### Expected behavior

As "Make a copy" is available from the drive, it should also be possible to do it from the document itself.

### What opreating system are you using?

Apple macOS

Mozilla Firefox

### Extensions

uBlock Origin, Wappalyzer, Passbolt

### Version

5.3.0

lemondevxyz:
Fixes #1061

This commit adds a 'Make A Copy' to all onlyoffice applications theortically. Though, I only tested sheets since that's all I had in my toolset.

The toolbar works through adding buttons with 'createButton' in `./www/common/common-ui-elements.js:569` and there are (thankfully) premade buttons. That means that buttons like 'Export', 'Make A Copy', 'Import', etc can be created by supplying the type and a boolean indicating the right side.

But most uses of createButton are through
`./www/common/sframe-common.js:106` which is the same underlying function of `common-ui-elements.js` but with the first two arguments skipped.

Anyhoo, this was long. Basically, all this commit does is add two lines at `./www/common/onlyoffice/inner.js:2851`.

### DIFF
--- a/www/common/onlyoffice/inner.js
+++ b/www/common/onlyoffice/inner.js
@@ -2849,6 +2849,9 @@ Uncaught TypeError: Cannot read property 'calculatedType' of null
 
             var $properties = common.createButton('properties', true);
             toolbar.$drawer.append($properties);
+            
+            var $copy = common.createButton('copy', true);
+            toolbar.$drawer.append($copy);
         };
 
         var noCache = false; // Prevent reload loops
--- a/www/common/sframe-common-outer.js
+++ b/www/common/sframe-common-outer.js
@@ -1350,7 +1350,7 @@ define([
                     channel: secret.channel,
                     href: currentPad.href,
                     password: password,
-                    title: currentTitle + " (copy)",
+                    title: currentTitle
                 };
                 var obj = { d: data };
                 var href = window.location.pathname;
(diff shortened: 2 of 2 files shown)


wginolas:
Thanks for your help @lemondevxyz !

I tested your change and it works fine. I, however, found a related bug we can maybe also resolve in this PR: When copying something in CryptPad usually " (copy)" is appended to the document name. This should also happen with OnlyOffice documents.

lemondevxyz:
@wginolas I added a new commit and it fixes that issue you were writing about it.

wginolas:
Hi @lemondevxyz, adding "(copy)" to the document name is more complicated than I thought. We have to make sure this works correctly for all apps, drive and also for all languages. I will create a dedicated bug report for this.

Can you remove the last change from this PR? I will merge it then.

lemondevxyz:
Oops, I wasn't thorough with testing the copied file title. I reverted the changed.

PR #1160, opened 2023-07-22
set cursor in OTP input field

ansuz:
this is just a minor usability improvement - the user's cursor is automatically set inside the OTP input field when they are prompted. Otherwise they have to select the field manually.


ansuz:
I noticed that the settings page could also use the same automatic focus on the OTP field, as well as keyup listeners to submit OTPs and passwords.

wginolas (approved):
Tested and reviewed. Looks good to me. Thanks, @ansuz !

PR #1166, opened 2023-07-27
fix 1password generating passwords on login page

### LINKED ISSUE #1164: Replace "Sign up" button on login page with a link
Feedback from a [Mastodon thread](https://hachyderm.io/@tobiaalberti/110758089522415967)

The "Sign up" button on the login page creates confusion because it implies that if pressed it will create a new account with the supplied credentials. The presence of this button could also be the source of issues with One Password not recognizing the page, although we have not reproduced this and it seems unlikely

To avoid confusion we could replace the "Sign up" button with a link to the register page and the text: "Don't have an account? Sign up"

lemondevxyz:
Fixes #1164 

Basically, the login page has a button named "Sign Up". Whenever 1password, and possibly other password managers, see a button with a "Sign Up" text, they generate a password instead of inputting an already existing one.

This fix is simple:
1. It changes the button in the login page from a button to a link
2. It changes the styling of the link to look like the old one.

Tested on Google Chrome, Linux, with the 1password extension.

https://github.com/cryptpad/cryptpad/assets/62130041/0d7cec5a-0da9-4897-8004-647cd11e9057

---

Note: The staging branch was not working for this fix, as in I literally couldn't run `npm run start`, so I switched to the main branch.


davidbenque:
Thanks @lemondevxyz ! 
I think I would prefer if the link was just text rather than styled like a button. I am focused on wrapping up 5.4.0 right now and then will be off for a week. But I'll come back to this afterwards.

Issue #1162, opened 2023-07-25
Settings sidebar not accessible


dariiing:
Sidebar is not accessible for keyboard users

PR #1168, opened 2023-07-31
Make Elements accessible using keyboard (issue 1162)

lemondevxyz:
This PR request basically fixes issue #1162. It tackles every step in the issue, namely:
- Settings Page sidebar
- Teams page (content and sidebar)
- User list (only documentation was visible before this PR)
- Contacts list

Tested on Google Chrome, Desktop, Linux.

Do note that the fixes are in several commits, be sure to check the commits' description for more information.

### DIFF
--- a/customize.dist/src/less2/include/drive.less
+++ b/customize.dist/src/less2/include/drive.less
@@ -128,7 +128,7 @@
         }
     }
 
-    div:focus {
+    div:not(:focus-visible):focus {
         outline: none;
     }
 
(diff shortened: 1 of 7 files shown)


wginolas:
Thanks lemondevxyz for the PR! In my opinion we should find a solution with less code duplication. @yflory will be available in a week. We can discuss this topic then.

lemondevxyz:
I could see an application wide solution being made through wrapping jQuery's `.click` with an event handles that handles both mouse clicks and the enter key.

Or, a solution could be in the form of a button component similar to the `UIElements.createButton` function in `./www/common/common-ui-elements.js`. If no code refactoring is needed one could simply scan all elements using jQuery after the document has loaded and modify all elements with a `onclick` event to add all the necessary attributes to make those elements keyboard selectable.

PR #1173, opened 2023-08-06
prototype protestware against google's proposed WEI API

ansuz:
# :wave: 

This is only intended as a prototype and will require further work before merging (if you decide to merge).

The idea is to:

1. detect the presence of Google's proposed Web Environment Integrity API
2. stop CryptPad from loading if it is detected
3. briefly explain why it's bad
4. propose that the user install a browser like Firefox
5. point them towards a (still non-existent) blog post where they can learn more


ansuz:
I pushed another (less invasive) commit to measure how widely the API is supported.

...and I realized a screenshot might be helpful:

image

PR #1187, opened 2023-08-20
OO disallow " (copy)" from being added to files used with template

### LINKED ISSUE #1177: "(copy)" is added to OnlyOffice document title when creating from a template
1. Create a new Sheet, Document or Presentation
2. Select a template at the creation screen
3. Note that the document title is the template name + " (copy)"

This is not the case when creating other document types so the "(copy)" should be removed for consistency.

lemondevxyz:
Closes #1177

Pull Request #1067 added " (copy)" to copied OnlyOffice files. But, it made a mistake in the sense that it added it to all files that use a template. (copied files use template internally).

This pull request allows users to create a file using a template without " (copy)" being added.

### DIFF
--- a/www/common/onlyoffice/inner.js
+++ b/www/common/onlyoffice/inner.js
@@ -235,13 +235,14 @@ define([
 
 
         var getFileType = function () {
-            var type = common.getMetadataMgr().getPrivateData().ooType;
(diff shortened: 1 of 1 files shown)


wginolas:
Successfully tested this PR.

Issue #604, opened 2020-09-01
example nginx config drops security headers in nested sections


martinetd:
Running a linter such as `gixy` on the default config yields warnings that headers are dropped with add_header:

```
>> Problem: [add_header_redefinition] Nested "add_header" drops parent headers.
Description: "add_header" replaces ALL parent headers. See documentation: http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header
Additional info: https://github.com/yandex/gixy/blob/master/docs/en/plugins/addheaderredefinition.md
Reason: Parent headers "x-xss-protection", "x-content-type-options", "content-security-policy" was dropped in current level
```

This is a really sneaky problem, please fix the example config (unfortunately I'm not aware of any "nice" way of fixing this without either splitting headers in another file or using an extension, which is why I'm posting an issue rather than a PR :/)

(BTW while I'm nitpicking about the nginx config, my initial nginx config was just proxying both domains to ports 3000 and 3001 but the comments in config.example.js are wrong about the port httpSafePort. it doesn't default to other port +1, it just doesn't listen if not set - i had to uncomment it to work)

djmattyg007:
Personally I think it would be ideal to have separate server blocks entirely for each domain.

spinus:
> Personally I think it would be ideal to have separate server blocks entirely for each domain.

Even if the split is made, question remains. What headers should be set in both cases?

mthld:
Hello,

I invite you to take these configuration discussions over our community forum: https://forum.cryptpad.org

Closing this now. :)

PR #1213, opened 2023-08-26
Fix nginx example config

martinetd:
- remove stray '^' char
- fix gixy warnings regarding add_header: add_header in a sublevel drops all headers set at parent level, see this link for details: https://github.com/yandex/gixy/blob/master/docs/en/plugins/addheaderredefinition.md
  * /api/ location was setting CORP/COEP to the same value as server level, just drop the add_header calls there
  * blob/block OPTIONS was missing dropping some headers, repeat the ones from top level

Fixes: #604


martinetd:
> Note that the stray '^' is already fixed in 5.4.1

Ah, figured it couldn't be broken for long but I probably should have checked :)

I've rebased and remved this part.

martinetd:
Oh, so requests going to node should either add a bunch of proxy_hide_header, or the behaviour of dropping some headers from nginx was actually useful... I wonder if there's a way to explain that to gixy.. Well, won't be a problem with the new config anyway, thanks.

I'll test a basic nginx config as I don't need any performance on my end (feeling it's a bit of a shame to not at least serve blocks and the dozens of js files directly, but won't matter much to me)
Just to be clear, it basically skips all the alternative backend ports (safe origin port and ws port) and should just work anyway?

PR #1258, opened 2023-09-29
Update .gitignore: Add package-lock.json

FloGatt:
Since we run npm install during installation [0], we need to add package-lock.json to .gitignore.

[0] https://docs.cryptpad.org/en/admin_guide/installation.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ block/
 logs/
 privileged.conf
 config/config.js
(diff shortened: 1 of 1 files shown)


wginolas:
Hi @FloGatt! Thanks for your contribution. After discussing this with the team, we decided to keep `package-lock.json` in the repository and update the installation guide instead.

Issue #1171, opened 2023-08-02
Add a description field to an event


ghost:
As requested by some of our users, it would be convenient to have a "Description" field for adding more information to an event. It could sit between "Title" and "Location".

PR #1299, opened 2023-10-20
Feature: Add a description field for calendar events

Chouhartem:
- Add a description field when creating and editing an event as per #1171 
  - This field supports markdown information to format the description.
- Integration in the detailed popup view of [tui.calendar](https://github.com/nhn/tui.calendar).
- Support for plaintext/markdown export
  - html export is not supported yet (https://www.rfc-editor.org/rfc/rfc5545#section-3.2.1)

Note: Two new descriptions keys : `calendar_desc` and `calendar_description` are to be added upon version release.

### DIFF
--- a/www/calendar/app-calendar.less
+++ b/www/calendar/app-calendar.less
@@ -197,7 +197,9 @@
                     background-color: @cp_dropdown-bg-hover;
                 }
                 .tui-full-calendar-content {
-                    white-space: nowrap;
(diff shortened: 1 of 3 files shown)


Chouhartem:
Hello,

I rebased with staging, integrated the changes that have been made and add the following features:
* Use Codemirror to edit the description field in a nice way.
* Fix a bug when editing a recurring event

PR #1327, opened 2023-11-20
Cleanup repo from outdated/misplaced files

mthld:
Here is a proposal to remove outdated files from the repository.

It'll allow us to keep things a bit more tidy and add clarity to the repo by relying more on the [documentation](https://docs.cryptpad.org) and less on `Readme.md` files hidden in source code directories.


davidbenque:
some comments on the very different things listed

- [/www/bounce/readme.md](https://github.com/cryptpad/cryptpad/blob/staging/www/bounce/readme.md)
  - imo this should be kept to document what this app is
- [/www/lib/ical/README.md](https://github.com/cryptpad/cryptpad/blob/staging/www/lib/ical/README.md)
  - `www/lib/ical/` could be removed as the licensing info and license are now provided by headers and the `LICENSES` folder
  - The information in `www/lib/changelog.md` about this dependency is not consistent with the info in the readme. IMO we should keep the relevant info in the changelog file
- [/www/pad/wordcount/README.md](https://github.com/cryptpad/cryptpad/blob/staging/www/pad/wordcount/README.md) probably needs to be kept 
- [/customize.dist/src/less2/readme.md](https://github.com/cryptpad/cryptpad/blob/staging/customize.dist/src/less2/readme.md) needs to be kept as it documents our use of LESS
- [/customize.dist/fonts/cptools/README.md](https://github.com/cryptpad/cryptpad/blob/staging/customize.dist/fonts/cptools/README.md) provides licensing info for 1 icon, so needs to be kept

davidbenque:
I think that removing the files below will reduce the discoverability of information. So we could keep the readme files with a link to the place where the information has been moved

- [/www/common/translations/README.md](https://github.com/cryptpad/cryptpad/blob/staging/www/common/translations/README.md)
- [/customize.dist/readme.md](https://github.com/cryptpad/cryptpad/blob/staging/customize.dist/readme.md)
- [/customize.dist/translations/README.md](https://github.com/cryptpad/cryptpad/blob/staging/customize.dist/translations/README.md)

Issue #973, opened 2022-10-14
CryptDrive can't be downloaded


dullbananas:
### Contribution guidelines

- [X] I've read the [code of conduct](https://github.com/xwiki-labs/cryptpad/blob/main/CODE_OF_CONDUCT.md) and wholeheartedly agree

### I've found a bug and checked that ...

- [X] I understand that not following the below instructions will result in immediate closure and/or deletion of my issue.
- [X] I have understood that this bug report is dedicated for bugs, and not for support-related inquiries.
- [X] I have understood that answers are voluntary and community-driven, and not commercial support.
- [X] I have verified that my issue has not been already answered in the past. I've read the [Common issues documentation section](https://docs.cryptpad.org/en/FAQ.html#common-issues) and I also checked [previous issues](https://github.com/xwiki-labs/cryptpad/issues).

### Description

The download button for downloading the entire CryptDrive does nothing.

### Steps to reproduce

1. Go to the CryptDrive section in settings
2. Press "download my cryptdrive" button, then "ok"
3. Wait for download and decryption
4. Press "download"

### Expected behavior

The zip file is downloaded

### System information

| Question | Answer |
| --- | --- |
| Operating system and version | iOS 16.0.3 |
| Web browser and version| Safari |
| Extensions installed | None |
| Browser tweaks | None |
| CryptPad version | 5.1.0 |

mthld:
Hello!

Thanks for the report, I can surely reproduce this issue. The UI breaks on my iPhone SE.

https://user-images.githubusercontent.com/41020854/196118719-525635fc-0eff-415b-8d5b-1af44f1b6220.mp4

You didn't mentioned the device you are using, can you tell us please? :)

mthld:
I just tried on an Android smartphone and it works as expected over here.

mthld:
Here is the logs I extracted from the Safari console:
```
[Error] Cancelled load to blob:https://sandbox.cryptpad.info/58e9650d-bc69-4291-ba4f-38a3d8ed2a9b because it violates the resource's Cross-Origin-Resource-Policy response header.
[Error] Failed to load resource: Cancelled load to blob:https://sandbox.cryptpad.info/58e9650d-bc69-4291-ba4f-38a3d8ed2a9b because it violates the resource's Cross-Origin-Resource-Policy response header. (58e9650d-bc69-4291-ba4f-38a3d8ed2a9b, line 0)
```

dullbananas:
[...]

PR #1514, opened 2024-06-17
Make httpSafePort work as expected when httpSafeOrigin is set

### LINKED ISSUE #907: httpSafePort is not loaded from config when httpSafeOrigin is set
**Describe the bug**
When `httpSafeOrigin` is set in config.js, `httpSafePort` is not initialized properly in env.js.

**Where did it happen?**
I am packaging CryptPad for Cloudron. Found this when making the package.

**To Reproduce**
Have a config.js like below:
```
    httpSafeOrigin: `https://some.domain.com`,
    httpSafePort: 3001,
```

CryptPad does not listen on port 3001 as expected from above config.

**Expected behavior**
`httpSafePort` must be loaded from config.

dullbananas:
Fixes #907

Required for using a separate sandbox origin in a remote development environment, which I need for reproducing #973


dullbananas:
I'm using a remote development environment where the URL used in the local browser to access the ports is more like 3000.example.com and 3001.example.com

dullbananas:
That's not true. Setting httpSafeOrigin causes the server to not use port 3001, which causes 3001.example.com to not exist.

PR #1516, opened 2024-06-18
Switch to new `http2` Nginx option

ghost:
As we promote and support Nginx stable, every release since v1.26.0 needs to use a new dedicated option for `http2`.

Both example files have been updated:
- `example.nginx.conf`
- `example-advanced.nginx.conf`

### DIFF
--- a/docs/example-advanced.nginx.conf
+++ b/docs/example-advanced.nginx.conf
@@ -9,11 +9,13 @@
 #   in production and require professional support please contact sales@cryptpad.fr
 
 server {
-    listen 443 ssl http2;
-    listen [::]:443 ssl http2;
+    listen 443 ssl;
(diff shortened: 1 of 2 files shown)


nderambure:
Note that on last Debian stable (12), nginx is 1.22.1 so this default config does not work and must be written like previously.

nderambure:
Thx @mathilde-cryptpad for your answer.
I should have check the nginx stable version, I was naïvely thinking that Debian choosed nginx as a stable version, not legacy, and was conforted by the fact that nginx support websockets since 1.4 version since I remember correctly.

What could be nice is to add an info in the doc saying : "Note that current Debian stable uses nginx legacy version, you must install nginx stable from official repository.".

PR #1547, opened 2024-07-05
Insert onlyoffice api

wginolas:
Sorry for this not very readable change. I reorganized some code. I also tested, if OnlyOffice still works inside Nextcloud.


wginolas:
Not ready yet

Issue #1099, opened 2023-07-05
Performance table not responsive for mobile


wginolas:
### Description

![image](https://github.com/cryptpad/cryptpad/assets/1014026/560b60ff-bf69-4057-99dd-6f9326f8757a)

### Steps to reproduce

Open the performance page in the admin tools

### Expected behavior

The table should fit on the screen.

### What opreating system are you using?

Google Android

### What browsers are you seeing the problem on?

Mozilla Firefox

### Extensions

.

### Version

5.3.0

PR #1555, opened 2024-07-11
Make performance table responsive for mobile

DianaXWiki:


davidbenque:
Addressing #1099

davidbenque (changes requested):
Some comments about the code. Otherwise the table layout seems to work well, please generalize it to all tables in the admin app

davidbenque:
This looks good, all issues seem addressed. Remaining question is about code organisation, right now we have:
- `.cp-sidebar-table` in sidebar-layout.less
- `#cp-admin-table` in app-admin.less
But these are always used together and it seems like we would want any new tables in a sidebar-layout to benefit from the horizontal scrolling on mobile. So before merging we should check where this code should be and potentially bring all table styles to sidebar-layout.less

davidbenque (approved):

Issue #1478, opened 2024-05-01
Print menu item changes to document text after cancelling print


lucidcocoon:
### Description

I am able to get the "Print" menu item to change to the content of my Code markdown file simply by cancelling a print. The print button is still functional but it will not read "Print" again until I reload the page. Tested this in the latest Firefox on Opensuse Aeon.

In the following screenshot the markdown file contains the following:
```
This is a test
```

![Screenshot 2024-05-02 at 00-04-36 Encrypted Code](https://github.com/cryptpad/cryptpad/assets/168688782/d5aee5c4-f97a-4393-8556-9d6b8fc4ad09)

I should also note that if I inspect the source code using the Firefox inspector tool, the code changes from this:
```
<li role="menuitem" tabindex="0">
  <a class="fa fa-print cp-toolbar-icon-print cp-toolbar-rightside-button" tabindex="-1" data-tippy="" data-original-title="Print your document or export it as a PDF file">
    <i class="undefined"></i>
    <span>Print</span>
  </a>
</li>
```

to this:

```
<li role="menuitem" tabindex="0">
  <p>This is a test</p>
</li>
```

### Steps to reproduce

1. Load https://cryptpad.fr/ as a guest and click Code
2. Add some content to the file in the code editor pane
3. Click on File > Print
4. Click *cancel* on the print dialog
5. Click on the File menu again

### Expected behavior

The Print menu item should not have the content changed.

### Which operating system are you using?

Linux/BSD/UNIX

### What browsers are you seeing the problem on?

Mozilla Firefox

### Extensions

None

### Version

2024.3.0

PR #1557, opened 2024-07-11
Fix overlapping variable names causing issues when printing Code documents

DianaXWiki:


DianaXWiki:
This PR should also fix [#1478](https://github.com/cryptpad/cryptpad/issues/1478)
The content instead of being displayed in `Preview`  for printing was appended to the dropdown menu instead of the actual button due to a variable name overlapping.

image

yflory (approved):

Issue #1522, opened 2024-06-21
Lending ownership before sharing a password-protected document creates a bogus file


Chouhartem:
### Description

When adding an extra owner who doesn’t have access to a password-protected document, it send them a notification that adds the document to their drive upon acceptance:

![ownership validation](https://github.com/cryptpad/cryptpad/assets/10630640/11498031-ec9f-4f38-ac5a-491233f00c39)

Clicking on “Open the document in a new tab” asks for the password and then accepting it will work (as the document has already been decrypted once), and all is well in the best of possible worlds.

However, if the to-be owner accepts it without opening the document first, the owned document is added to their drive in an unopenable state:

![non-valid link](https://github.com/cryptpad/cryptpad/assets/10630640/54335150-a019-4674-8a13-7fd97db60ad5)

It’s not that critical as opening the document afterward, either via a share-link or the “Share to contact” function unlock this states, but it adds an extra step and it’s not the most intuitive.

### Steps to reproduce

User 1:
1. Create a new pad with a password on creation (or adding one afterward)
2. 🔓 Access > Owners > Add an owner from the contact.

User 2:
1. Receive the ownership request and accept it.
2. Try to open the new document from the drive.

### Expected behavior

The created file should be the same as the one in the “Open the document in a new tab” link on the invitation to allow decrypting the file after acceptance, or ask for password upon acceptance to store it in the drive as it does with other files.

### Which operating system are you using?

Linux/BSD/UNIX

### What browsers are you seeing the problem on?

Mozilla Firefox, Google Chrome

### Extensions

Not relevant.

### Version

2024.3.1

Issue #1521, opened 2024-06-19
Changing password prevents previous owners from accessing new copy


thejoelpatrol:
### Description

Changing the password of a document should just require users who previously accessed it to use the new password. Instead, some can never access it again because it permanently appears "destroyed" for them.

### Steps to reproduce

1. Create a new spreadsheet owned by you
2. Add another user as an owner (e.g. a contact)
3. Second user accepts ownership
4. Second user adds file to their cryptdrive
5. Document creator changes password
6. Second user attempts to access document via its share link, via their drive, or via a notification that it has been shared to them from a contact
7. Second user receives the error "This document was destroyed by an owner"
8. Document creator re-sends new share link, re-shares to contact, or re-adds second user as an owner: no change
9. Guest users can view the share link with the new password. Only the second owner who previously added it to their drive sees it as "destroyed". 

NOTE it *is* resolved by:
10. Second owner moves it to their trash, *and empties trash*, and *then* accesses the link with new password

### Expected behavior

The second user should be prompted for the new password and able to access the document. 

### Which operating system are you using?

Apple macOS

### What browsers are you seeing the problem on?

Mozilla Firefox, Google Chrome

### Extensions

Firefox: none

Chrome: GoFullPage, JSON Viewer, Privacy Badger, Save to Pocket, Adblock Plus

same behavior in both Chrome and Firefox

### Version

2024.3.1

thejoelpatrol:
A note on version: this occurred on the cryptpad.fr instance, so I'm not sure what version is running there, but I assume the most recent version.

Issue #1270, opened 2023-10-04
Sharing password protected document with contacts have you typing the password


ghost:
### Description

When sharing a password protected document to a contact, they shouldn't have to type the password to open the document.

### Steps to reproduce

1. Create a password protected document
2. Share it with one of your contacts
3. Your contact have to type in the password before accessing the document

### Expected behavior

The contact shouldn't have to type the password before accessing the document

### What opreating system are you using?

Linux/BSD/UNIX, Apple macOS

### What browsers are you seeing the problem on?

Mozilla Firefox

### Extensions

uBlock Origin, Wappalyzer, Passbolt

### Version

5.4.1

PR #1565, opened 2024-07-15
Fix ownership/sharing issues with password-protected pads

Chouhartem:
This PR fixes the following issues:
- [x] #1270 
- [x] #1521
- [x] #1522 

Moreover it also fixes an UI error when the preview link when sharing the ownership of a document via CryptPad is broken.

Extra change: make the `block hash` accessible to sharedWorkers.

### DIFF
--- a/www/common/sframe-common-outer.js
+++ b/www/common/sframe-common-outer.js
@@ -1083,7 +1083,7 @@ define([
                         Cryptpad.addSharedFolder(null, secret, cb);
                     } else {
                         var _data = {
-                            password: data.password,
+                            password: data.pw || data.password,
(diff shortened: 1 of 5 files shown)


www/common/sframe-common-outer.js
```diff
@@ -1083,7 +1083,7 @@ define([
                         Cryptpad.addSharedFolder(null, secret, cb);
                     } else {
                         var _data = {
-                            password: data.password,
+                            password: data.pw || data.password,
```
Chouhartem, line 1086: I don’t think the `|| data.password` part is necessary.

yflory (approved):

PR #1569, opened 2024-07-17
Make password change confirmation button responsive on mobile

DianaXWiki:


customize.dist/src/less2/include/alertify.less
```diff
@@ -327,6 +327,8 @@
             nav {
                 padding: @alertify_padding-base;
                 text-align: right;
+                display: flex;
+                align-items: center;
```
davidbenque, line 331: The level of this change is way too high. All `nav` in modals are affected, check for example the Share modal

davidbenque (changes requested):

customize.dist/src/less2/pages/page-install.less
davidbenque, line 1: in case `.btn-danger` was used elsewhere in page-install, it now does not get styles that it used to. To check that the confirm button is indeed the only button that this was used for

customize.dist/src/less2/pages/page-register.less
davidbenque, line 1: in case `.btn-danger` was used elsewhere in page-install, it now does not get styles that it used to. To check that the confirm button is indeed the only button that this was used for

www/common/common-interface.js
```diff
@@ -775,10 +775,10 @@ define([
         });
     };
 
-    UI.confirm = function (msg, cb, opt, force) {
+    UI.confirm = function (msg, cb, opt, force, className) {
         cb = cb || function () {};
         opt = opt || {};
-
+        console.log('Parameters:', { msg, cb, opt, force, className });
```
davidbenque, line 781: logging should be removed

www/settings/app-settings.less
```diff
@@ -38,6 +38,22 @@
         margin-left: 0.5rem;
     }
 
+    .alertify {
+        nav {
```
davidbenque, line 42: This affects all buttons navs in modals in settings. potentially some unintended consequences

www/settings/app-settings.less
```diff
@@ -113,6 +129,11 @@
                     margin-left: 10px;
                 }
             }
+            nav {
```
davidbenque, line 132: check scoping

davidbenque (approved):
Pending some checks on scoping (see comments) and testing of the install page confirm button which I did not do.

davidbenque:
image

The buttons should be aligned to the right side here. This used to be done with `text-align` but could also be done with a flex propoerty

Chouhartem:
Small screens in landscape (on firefox responsive design mode) don’t trigger the `@media` and let the button overflow:

screenshot_responsive

davidbenque:
👍🏻 this looks better from the UI side

yflory (approved):

PR #1571, opened 2024-07-19
env.js: fix httpSafePort handling

martinetd:
This is a follow-up of #1212 I had opened last year:

- The httpSafePort config item didn't work at all even if httpSafeOrigin wasn't set, so make it work.
- Since then the comment in the config has been clarified in #1518, but it still isn't clear enough as seen in #1514 -- improve the comment a bit.
- While here also make it clear that the websocket port is not optional
- ... and remove unused NO_SANDBOX variable in env.js that was adjacent to the httpSafePort setting; this probably doesn't really belong here but cannot hurt.

all in all there is nothing impacting any production setup, but might as well finish this so I can get it off my todo. Thanks!

### DIFF
--- a/config/config.example.js
+++ b/config/config.example.js
@@ -104,7 +104,7 @@ module.exports = {
  *
  *  In production environments, your reverse proxy (usually NGINX)
  *  will need to forward websocket traffic (/cryptpad_websocket)
- *  to this port.
+ *  to this port - it cannot be disabled.
  *
  */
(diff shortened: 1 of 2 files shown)


martinetd:
(rebased to fix conflict in 2025.9.0 -- this really should have zero impact on production and just fixes an optional setting that doesn't work right now. Perhaps it'd be better to remove the httpSafePort setting altogether instead?)

PR #1578, opened 2024-07-29
Misaligned file menu items

DianaXWiki:


www/common/common-ui-elements.js
```diff
@@ -638,34 +644,28 @@ define([
...
-                        accept: data ? data.accept : undefined,
-                        binary: data ? data.binary : undefined
-                    });
-
-                    var handler = data.first? function () {
-                        data.first(function () {
-                            importer(); // Make sure we don't pass arguments to importer
-                        });
-                    }: importer; //importContent;
+                button = createIconButton('fa fa-upload cp-toolbar-icon-import', '', Messages.importButtonTitle, Messages.importButton);
```
davidbenque, line 647: In doubt assign only the icon classes to the <i> element (`fa`, `fa-icon`, `cptools`, `cpools-icon`) and leave the other classes such as `cp-toolbar-icon-print` on the <a> container

davidbenque:
Approved on the design side but refactoring should be reviewed

DianaXWiki:
Note: there is still need to find the cause of the missing 'Help' menu item in the Diagram dropdown

www/common/common-ui-elements.js
```diff
@@ -638,34 +644,28 @@ define([
...
-                        accept: data ? data.accept : undefined,
-                        binary: data ? data.binary : undefined
-                    });
-
-                    var handler = data.first? function () {
-                        data.first(function () {
-                            importer(); // Make sure we don't pass arguments to importer
-                        });
-                    }: importer; //importContent;
+                button = createMenuItemButton('fa fa-upload', ' cp-toolbar-icon-import', Messages.importButtonTitle, Messages.importButton);
```
Chouhartem, line 647: typo: extra space in `'cp-toolbar-icon-import'`

Chouhartem (commented):
The refactor seems ok for me up to some minor fixes.

Chouhartem (approved):

www/common/common-ui-elements.js
```diff
@@ -623,13 +623,19 @@ define([
...
         if (!callback && data.callback) { callback = data.callback; }
+
+        function createMenuItemButton(iconClasses, buttonClasses, title, text, ariaLabel) {
+            return $('<button>', {
+                'class': buttonClasses,
+                'title': title,
+                'aria-label': ariaLabel
+            }).append($('<i>', {'class': iconClasses}))
+              .append($('<span>', {'class': 'cp-toolbar-drawer-element'}).text(text));
+        }
```
yflory, line 634: We almost never use named functions in the rest of the product, so we should probably use an anonymous function here too (using `let createMenuItemButton = function (...) {};` syntax).
We also try to avoid using jQuery whenever we can and this is a good example of a place where we can use hyperscript (the `h` function) instead.

About the content, I think we should only append the `<i>` tag if 'iconClasses' is not empty and the `<span>` tag if 'text' is not empty.
And I also think the "ariaLabel" argument is not necessary. Its value should probably always be set to title and fallback to text.

www/common/common-ui-elements.js
```diff
@@ -970,22 +948,16 @@ define([
                 });
                 break;
             case 'save': // OnlyOffice save
-                button = $('<button>', {
-                    'class': 'fa fa-save',
-                    title: Messages.settings_save,
-                }).append($('<span>', {'class': 'cp-toolbar-drawer-element'})
-                .text(Messages.settings_save))
+                button = createMenuItemButton('fa fa-save', '', Messages.settings_save, Messages.settings_save);
```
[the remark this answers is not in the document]
yflory: This is only visible in dev mode so not a big issue, a fix would require a change to the style to avoid applying a margin-right on buttons without text.

yflory (changes requested):

www/common/common-ui-elements.js
```diff
@@ -623,13 +623,19 @@ define([
         var appType = (common.getMetadataMgr().getMetadata().type || 'pad').toUpperCase();
         data = data || {};
         if (!callback && data.callback) { callback = data.callback; }
+
+        function createIconButton(iconClasses, buttonClasses, title, text, ariaLabel) {
```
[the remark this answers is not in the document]
yflory: The new name "createMenuItemButton" doesn't seem right to me either because some button are not menu entries.
What this function does it to create a base button that can be customized for each use-case below. Maybe just "makeButton"?

yflory (approved):

PR #1583, opened 2024-08-03
Make 'New Event' modal appear in user view for small screens

DianaXWiki:


davidbenque (approved):
UI approved

www/calendar/app-calendar.less
```diff
@@ -155,6 +155,10 @@
         }
         .tui-full-calendar-popup {
             border-radius: @variables_radius_L;
+            @media screen and (max-width : @browser_media-medium-screen){
+                position: fixed;
+                top: 5.5rem;
```
yflory, line 160: Using `positon: fixed` without a max-height is risky:
image
In this example, I can't reach the "Save" button because it is hidden outside of the screen and since the position is fixed, scrolling doesn't change anything.
A fix could be to add a max-height to the modal and make it scrollable with the "overflow" property.

yflory (changes requested):

Issue #1371, opened 2024-01-08
Should be able to toggle multiple calendars in/out of view


zuzanna-maria:
### Description

When viewing Calendar app on mobile, if many calendars are present, this impedes being able to see the actual calendar grid.
![Screenshot 2024-01-08 at 13 20 53](https://github.com/cryptpad/cryptpad/assets/34193395/08dbb4be-cf73-49f2-a479-66fbc4d38e03)

### Steps to reproduce

1. Access Calendar on mobile
2. If many different calendars are present, the actual calendar grid is out of view and scrolling down is necessary

### Expected behavior

There should be a way to toggle the visibility of multiple calendars to make viewing the calendar grid easier.

### Extensions

n/a

### Version

5.6.0

PR #1584, opened 2024-08-03
Enable toggle in and out of calendars on small screens

DianaXWiki:


davidbenque:
It is unclear how/when the "Show Calendars" button is triggered.
- [ ] When manually decreasing the screen width (e.g. in Firefox responsive view) the button does not get added when moving to a small size
- [ ] When manually increasing the screen width the button stays in desktop mode but does not do anything when clicked

davidbenque:
- [ ] "Show Calendars" should be changed to "Hide Calendars" when the list is shown

davidbenque (commented):
This seems OK from the UI side. Needs a proper code review

www/calendar/inner.js
```diff
@@ -818,21 +818,37 @@ define([
             });
         }
         if (APP.$calendars) { APP.$calendars.append(calendar); }
-        return calendar;
+        return $calendar;  // return jQuery element
+    };
+
+    var appendCalendarEntries = function (teamId, filter) {
+        var calendars = filter(teamId);
+        var $entriesContainer = $('<div class="cp-calendar-entries"></div>');
```
yflory, line 826: We should try to use hyperscript (the `h` function) and avoid jQuery as much as possible, especially when we want to make mobile improvements.

www/calendar/inner.js
```diff
@@ -845,12 +861,13 @@ define([
...
+
             var tempCalendars = filter(0);
             if (tempCalendars.length && tempCalendars[0] === APP.currentCalendar) {
-                APP.$calendars.append(h('div.cp-calendar-team', [
+                var $tempCalendarTeam = $(h('div.cp-calendar-team', [
                     h('span', Messages.calendar_tempCalendar)
-                ]));
-                makeCalendarEntry(tempCalendars[0], 0);
+                ])).appendTo(APP.$calendars);
+                var $tempCalendarEntries = appendCalendarEntries(0, filter).appendTo(APP.$calendars);
```
yflory, line 870: I don't think "appendCalendarEntries" is a good name for this function. As far as I understand, this function creates a container for the calendar entries and fills it, but the name suggests that it is going to append these entries to an existing element.

www/calendar/inner.js
```diff
@@ -868,7 +885,7 @@ define([
                     });
                 });
                 if (APP.loggedIn) {
-                    APP.$calendars.append(h('div.cp-calendar-entry.cp-ghost', importTemp));
+                    $tempCalendarEntries.append(h('div.cp-calendar-entry.cp-ghost', importTemp));
```
yflory, line 888: The `$tempCalendarEntries` container is hidden by default so neither the calendar entry with its title and actions nor the "import" button are visible. It seems the goal was to display the "show calendars" button here but, in my opinion, it doesn't make sense to show this button when there is a single entry.

www/calendar/inner.js
```diff
@@ -907,15 +926,56 @@ define([
...
+                    createShowCalendarsButton(teamId, filter, $calendars);
+                } else {
+                    calendars.forEach(function (id) {
+                        makeCalendarEntry(id, 1);
+                    });
+                }
             });
+        }
+        function createShowCalendarsButton(teamId, filter, $parentContainer) {
+            var $showCalendarsContainer = $('<div class="cp-calendar-entry cp-ghost"></div>').appendTo($parentContainer);
```
yflory, line 941: Try to use hyperscript in these cases, even if you need a jQuery object:
`$(h('div.cp-calendar-entry.cp-ghost'))`

www/calendar/inner.js
```diff
@@ -907,15 +926,56 @@ define([
...
+            var showCalendarsBtn = h('button.cp-calendar-showcalendars', [
+                h('i.fa.fa-eye'),
+                h('span', Messages.calendar_show),
+                h('span')
```
yflory, line 945: Empty `span` to remove

yflory (changes requested):
The "temporary calendar" case is not working and was probably not tested: we don't see the calendar name and can't do anything with it.
In general, I don't think the way this issue has been handled is good enough:
* In the case where only one calendar is present, we're replacing this calendar with a button taking the same space. Clicking on this button **adds** the hidden entry, which makes this new system take the same space as 2 entries.
  * This means that, with 1 or 2 calendars in your account or a team, the old system was better
* If a user has many teams with calendars, we end up in the same state than what we wanted to fix in the PR (see #1371)

I think a better solution would be to partially or completely hide the list of calendar until the user clicks on a button. By "partially hiding", I'm thinking of something like the "teams" app where we only see the "icons" of the menu entries (the calendar colors in this case?) but it may also be better to keep the current menu and show or hide it with a button.

www/calendar/inner.js
```diff
@@ -818,106 +818,125 @@ define([
...
+                calendars.forEach((id) => {
+                    var calendarEntry = makeCalendarEntry(id, teamId);
+                    $contentContainer.append(calendarEntry);
                 });
             });
+            if(isMobileView) {
+            if (totalCalendars > 2) {
+                var $showContainer = $(h('div.cp-calendar-entry.cp-ghost')).appendTo($calendars);
+                var showCalendarsBtn = h('button', [
+                    h('i.fa.fa-eye'),
```
davidbenque, line 907: Please toggle this for `fa-eye-slash` when the text says "Hide calendars"

www/calendar/inner.js
```diff
@@ -818,106 +818,125 @@ define([
...
-            myCalendars.forEach(function (id) {
-                makeCalendarEntry(id, 1);
-            });
-
-            // Add new button
-            var $newContainer = $(h('div.cp-calendar-entry.cp-ghost')).appendTo($calendars);
+            // Add the new calendar button
+            var $newContainer = $(h('div.cp-calendar-entry.cp-ghost')).appendTo($contentContainer);
             var newButton = h('button', [
                 h('i.fa.fa-calendar-plus-o'),
```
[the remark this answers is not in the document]
davidbenque: This should be applied to all icons

www/calendar/inner.js
```diff
@@ -775,7 +775,7 @@ define([
         var data = APP.calendars[id];
         var edit;
         if (data.loading) {
-            edit = h('i.fa.fa-spinner.fa-spin');
+            edit = h('i.fa.fa-spinner.fa-spin', {'aria-hidden': 'true'});
```
yflory, line 778: This doesn't look related to the initial goal of the PR, which was to fix a UI issue on small screens.

www/calendar/inner.js
```diff
@@ -796,8 +796,8 @@ define([
                 h('i.cp-calendar-inactive.fa.fa-calendar-o')
             ]),
             h('span.cp-calendar-title', md.title),
-            data.restricted ? h('i.fa.fa-ban', {title: Messages.fm_restricted}) :
-                (isReadOnly(id, teamId) ? h('i.fa.fa-eye', {title: Messages.readonly}) : undefined),
+            data.restricted ? h('i.fa.fa-ban', {title: Messages.fm_restricted, 'aria-hidden': 'true'}) :
+                (isReadOnly(id, teamId) ? h('i.fa.fa-eye', {title: Messages.readonly, 'aria-hidden': 'true'}) : undefined),
```
yflory, line 800: Same as the previous comment, this doesn't seem related to the initial goal of the PR.

www/calendar/inner.js
```diff
@@ -818,106 +818,130 @@ define([
...
-                            console.error(err);
-                            return void UI.warn(Messages.error);
-                        }
-                    });
-                });
-                if (APP.loggedIn) {
-                    APP.$calendars.append(h('div.cp-calendar-entry.cp-ghost', importTemp));
-                }
-                return;
-            }
```
yflory, line 874: My previous review stated that
> The "temporary calendar" case is not working and was probably not tested

This change makes it so that the feature has been completely removed. I agree that it is indeed not broken anymore, but that's not what we want either !

www/calendar/inner.js
```diff
@@ -818,106 +818,130 @@ define([
...
+            var $contentContainer = $(h('div.cp-calendar-content')).appendTo($calendars);
+
             if (myCalendars.length) {
                 var user = metadataMgr.getUserData();
                 var avatar = h('span.cp-avatar');
                 var uid = user.uid;
                 var name = user.name || Messages.anonymous;
-                common.displayAvatar($(avatar), user.avatar, name, function(){}, uid);
-                APP.$calendars.append(h('div.cp-calendar-team', [
+                common.displayAvatar($(avatar), user.avatar, name, () => {}, uid);
```
yflory, line 865: Syntax change only

www/calendar/inner.js
```diff
@@ -818,106 +818,130 @@ define([
...
+                $contentContainer.append(h('div.cp-calendar-team', [
                     avatar,
-                    h('span.cp-name', {title: name}, name)
+                    h('span.cp-name', { title: name }, name)
```
yflory, line 868: SameSyntax change: should be remvoed from the PR

www/calendar/inner.js
```diff
@@ -818,106 +818,130 @@ define([
...
+                    h('span')
+                ]);
+                $contentContainer.append($teamContainer);
+                calendars.forEach((id) => {
+                    var calendarEntry = makeCalendarEntry(id, teamId);
+                    $contentContainer.append(calendarEntry);
                 });
             });
+            if(isMobileView) {
+            if (totalCalendars > 2) {
```
yflory, line 904: Indentation issues

yflory (changes requested):
This PR should be fixed and cleaned:
* Some elements from my previous review has been ignored
  * The temporary calendar feature, which allows you to view a calendar shared by link, was broken in the last review and has now been completely removed
  * Some smaller issues are still here
* A big part of the PR is about adding spaces, new lines and replacing classic functions by arrow functions. This should be removed from the PR to make the diff easier to read
* I'm not really sure it's important to change the view when the screen is resized but since it's done we might as well keep it (but I haven't tested it)
* I'm not sure why the PR adds some "aria" attributes to things not related to the initial issue

www/calendar/inner.js
DianaXWiki: Some of these change with aria and adding UI elements have been requested in a previous UI review

www/calendar/inner.js
DianaXWiki: Same as before, these changes have been requested in an UI review so that the icons are not accessible to screen readers

yflory (approved):

PR #1586, opened 2024-08-05
Make code history buttons not overlap on small screens

DianaXWiki:


davidbenque (commented):
- [x] the height property is not properly applied
- [x] a small margin should also be applied between the buttons

davidbenque (commented):
I know there are already some `!important` in this part of the code, but I would like to limit adding new ones as much as possible. In this case you could do one of 2 things

- Move the media queries to the bottom of the file. The reason you cannot override the styles for small screens is that the media queries are at the top of the file, so the CSS "cascade" finds the desktop styles and uses them. This could have been done for a specific reason so you should carefully check for consequences if you move the media-queries to the bottom.
- Use more specific selectors in the media-queries. This will give them priority over other styles when the screen width condition is met without using `!important`.

davidbenque (approved):

yflory (approved):

PR #1598, opened 2024-08-16
Fix emoji avatar handling and team name overflowing in sidebar teams

DianaXWiki:


davidbenque:
👍🏻 looks OK at first UI check. Needs review of code and use of regex

www/common/inner/common-mediatag.js
```diff
@@ -124,7 +124,17 @@ define([
         return ANIMALS[seed % ANIMALS.length] || '';
     };
 
+    const emojiWithZWJRegex = /(\p{Emoji_Presentation}|\p{Emoji_Modifier_Base})(\p{Emoji_Modifier})?(?:\u200D(\p{Emoji_Presentation}|\p{Emoji_Modifier_Base})(\p{Emoji_Modifier})?)*/gu;
```
yflory, line 127: When using complex regex such as this one, please add a comment about what it does :)
FYI it seems this regex doesn't work in some cases from https://www.unicode.org/Public/emoji/latest/emoji-zwj-sequences.txt (the emoji is split with the regex). e.g. 🧑🏿‍❤️‍💋‍🧑🏽

www/common/inner/common-mediatag.js
```diff
@@ -124,7 +124,17 @@ define([
...
     var getPrettyInitials = MT.getPrettyInitials = function (name) {
+        let matches = name.match(emojiWithZWJRegex);
+        if (matches && name.startsWith(matches[0])) {
+            return matches[0];
+        }
+        else {
+            name = name.replace(emojiWithZWJRegex, '');
+            name = name.replace(/[\uFE0F\u200D\u2060]/g, '');
```
yflory, line 135: Same here, a small comment about what this is supposed to do please

yflory (changes requested):

yflory (approved):

PR #1603, opened 2024-08-24
Add an example config for Caddy

PoneyClairDeLune:
Translated from NGINX configs. Should allow Caddy as a viable alternative of NGINX for mass deployments.

If this is merged, the [corresponding documentation](https://docs.cryptpad.org/en/admin_guide/installation.html#install-and-configure-nginx) will need to be updated.


PoneyClairDeLune:
I'll start working on this at my next opportunity.

PoneyClairDeLune:
Emails changed and the two example config files have been moved to `docs/community`.

PR #1605, opened 2024-08-26
Missing margin on Features page

### LINKED ISSUE #1604: Insufficient Spacing Around Card Components
On the Features page, the last card in the layout(Premium, second screenshot) is missing the expected margin(Registered, first screenshot)
Screen Shot 2024-08-26 at 13 57 05
Screen Shot 2024-08-26 at 13 57 10

dariiing:
Hello,
This PR fixes #1604, adding a missing margin between card components.

### DIFF
--- a/customize.dist/src/less2/pages/page-features.less
+++ b/customize.dist/src/less2/pages/page-features.less
@@ -112,7 +112,11 @@
     .cp-regis-user {
         @media (max-width:575px) {
             margin-top: 3em;
-            margin-bottom: 3em;
+        }
+    }
+    .cp-premium-user {
+        @media (max-width:575px) {
+            margin-top: 3em;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


customize.dist/src/less2/pages/page-features.less
```diff
@@ -112,6 +112,7 @@
     .cp-regis-user {
         @media (max-width:575px) {
             margin-top: 3em;
+            margin-bottom: 3em;
```
yflory, line 115: I recommend adding an extra `margin-top` to the 3rd card, if it exists, instead of a `margin-bottom` here.
The 3rd card is only visible on cryptpad.fr where premium subscription are available so here you're adding an unnecessary margin for all the other instances.

yflory (changes requested):

yflory (approved):

Issue #1660, opened 2024-09-23
Arrow Key Navigation Fails Inside Ctrl+E Modal


dariiing:
When the Ctrl+E modal is open, the arrow keys work outside of the modal but not within it.

Steps to reproduce:
1. Open the drive
2. Open the Ctrl+E modal
4.  Use the arrow keys (you'll see that they work outside the modal)

PR #1669, opened 2024-10-01
Disable arrow key navigation in the drive while modal is active

dariiing:
This PR disables the use of arrow keys inside the drive while a modal is open, fixing #1660.
Previously, arrow keys could still be used to navigate files in the background, which could lead to unintended behavior or conflicts when interacting with the modal. Now, the arrow keys are disabled inside the drive while a modal is opened.

### DIFF
--- a/www/common/common-ui-elements.js
+++ b/www/common/common-ui-elements.js
@@ -2514,6 +2514,7 @@ define([
             else if (e.which === 13) {
                 if ($container.find('.cp-icons-element-selected').length === 1) {
                     $container.find('.cp-icons-element-selected').click();
+                    return;
                 }
             }
         });
--- a/www/common/common-ui-elements.js
+++ b/www/common/common-ui-elements.js
@@ -2511,13 +2511,13 @@ define([
                     next();
                 }
                 return;
-              }
+            }
             if (e.which === 13) {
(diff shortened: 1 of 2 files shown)


Chouhartem (approved):

www/common/common-ui-elements.js
```diff
@@ -2510,13 +2510,11 @@ define([
...
-                return;
-            }
-            if (e.which === 13) {
+              }
+            else if (e.which === 13) {
                 if ($container.find('.cp-icons-element-selected').length === 1) {
                     $container.find('.cp-icons-element-selected').click();
                 }
-                return;
             }
```
yflory, line 2518: Try to avoid this type of changes that don't have any impact on the results but only on the code style. I recommend to always run a `git diff` before creating a commit in order to clean unnecessary changes :)

www/common/drive-ui.js
```diff
@@ -1060,6 +1063,7 @@ define([
                             $elements.index($selection.last()[0]);
             var length = $elements.length;
             if (length === 0) { return; }
+
```
yflory, line 1066: Same here, please remove this type of changes from the commit/PR

yflory (changes requested):

yflory (approved):

Issue #1644, opened 2024-09-06
Missing padding between elements inside Form Settings


dariiing:
The Form Settings modal lacks space between elements, making it hard for certain users to use it.
![image](https://github.com/user-attachments/assets/3e70d674-dce2-4a24-a675-c31af77d7d4b)

PR #1679, opened 2024-10-08
Form settings modal accessibility

zuzanna-maria:
* Fixes insufficient spacing between elements on form settings modal (#1644)


davidbenque:
This is very similar to #1670 so perhaps the two PRs could be merged

Issue #1455, opened 2024-04-04
Can't set password for Form from access panel


zuzanna-maria:
### Description

The access panel in form documents doesn't provide a way to set a password or to edit it once set. A password can only be set during document creation.

### Steps to reproduce

1. Create owned form document without password
2. Click on the access panel and try to set password

### Expected behavior

Ability to set document password from access panel

### Extensions

n/a

### Version

2024.3.0

davidbenque:
Note: this was intentionally disabled as changing passwords caused the loss of existing responses to the form (you would need to know their author's private key to re-encrypt them with the new form keys)

zuzanna-maria:
Suggested improvement/feature: a warning displayed during document creation, notifying user that unless they set a password during creation, it will not be possible to do so later.

PR #1690, opened 2024-10-21
Form password warning

zuzanna-maria:
* Addresses #1455 (adds a warning that passwords can only be set during document creation for Forms)


customize.dist/src/less2/include/creation.less
```diff
@@ -165,6 +165,7 @@
                 //margin: 10px 0;
                 min-height: 28px;
                 line-height: 28px;
+                flex-flow: column;
```
yflory, line 168: This line breaks the style when the users have templates stored in their drives:
image
image

It adds scrollbars and the arrow button to show the next template page is under the list instead of next to it

yflory (changes requested):

yflory (approved):

PR #1695, opened 2024-10-28
Links included in Drive exports

### LINKED ISSUE #942: Links aren't parts of drive exports
**Describe the bug**
As of today links created in CryptDrive aren't parts of exports.

**Where did it happen?**
CryptPad.fr

**To Reproduce**
1. Go to https://cryptpad.fr/drive/
2. Create a link
3. Download your CryptDrive
4. Find no link in the files

**Expected behavior**
Links should be exported too, maybe under `.txt` format?

zuzanna-maria:
- Links are now part of Drive exports (fixes #942)


yflory (changes requested):

yflory (approved):

www/common/make-backup.js
```diff
@@ -276,8 +284,21 @@ define([
                         }
                     }, 50);
                 };
+                var todoLink = function () {
+                    var opts = {
+                        binary: true,
+                    };
+                    var fileName = getUnique(sanitize(rawName), '.txt', existingNames);
+                    existingNames.push(fileName.toLowerCase());
+                    var content = new Blob([fData.href], { type : "text/html;charset=utf-8" });
```
Chouhartem, line 293: 1. To be consistent with the rest of the codebase, there shouldn’t be a space before colon (it is mostly use for ternaries).
2. The file is a text file without html information in it, its mime type should hence be set accordingly (`text/plain;charset=UTF-8`).
3. Finally, it would be better to add a “newline” character at the end of the URL, otherwise some command line scripting can behave in a unexpected way:
    - For 2 links A and B containing urls `a` and `b` respectively, downloading them gives us two files `A.txt` and `B.txt`. Running `parallel firefox <{A,B}.txt` will attempt to open the url `ab` without the newline, but indeed opens the two urls with a `\n`.

All in all, I suggest the following fix:
```js
var content = new Blob([fData.href, '\n'], { type: "text/plain;charset=utf-8" });
```

yflory (changes requested):
While testing this PR, I noticed that links inside shared folders are missing from the exports (whether you download the entire shared folder or only a sub-folder inside the shared folder)

www/common/make-backup.js
```diff
@@ -286,26 +307,33 @@ define([
...
+                staticData = ctx.sf[el].static;
                 var sfData = ctx.sf[el].metadata;
                 var sfName = getUnique(sanitize((sfData && sfData.title) || 'Folder'), '', existingNames);
                 existingNames.push(sfName.toLowerCase());
-                return void makeFolder(ctx, ctx.sf[el].root, zip.folder(sfName), ctx.sf[el].filesData);
+                return void makeFolder(ctx, ctx.sf[el].root, zip.folder(sfName), ctx.sf[el].filesData, staticData);
             }
+            var sData;
             var fData = fd[el];
+            sd ? sData = sd[el] : sData = undefined;
```
Chouhartem, line 330:
```js
const fData = fd[el];
const sData = sd ? sd[el] : undefined;
```

Would be more idiomatic

www/common/make-backup.js
```diff
@@ -330,6 +357,11 @@ define([
             sframeChan: sframeChan
         };
         var filesData = data.sharedFolderId && ctx.sf[data.sharedFolderId] ? ctx.sf[data.sharedFolderId].filesData : ctx.data.filesData;
+        var links = ctx.sf[data.sharedFolderId] && ctx.sf[data.sharedFolderId].static ? ctx.data.static && ctx.sf[data.sharedFolderId].static : ctx.data.static;
+
+        Object.keys(links).forEach(function(key) {
+            filesData[key] = links[key];
+        });
```
yflory, line 364: You could avoid merge the data (and risking a conflict with ids) if you provide `links` as a 5th argument to `makeFolder` 5 lines below (line 369 when I'm writing this review).

www/common/make-backup.js
```diff
@@ -286,26 +307,32 @@ define([
...
+                return void makeFolder(ctx, ctx.sf[el].root, zip.folder(sfName), ctx.sf[el].filesData, staticData);
             }
             var fData = fd[el];
+            var sData = sd ? sd[el] : undefined;
             if (fData) {
                 addFile(ctx, zip, fData, existingNames);
                 return;
+            }  else if (sData) {
+                addFile(ctx, zip, sData, existingNames);
+                return;
```
yflory, line 335: This is fine but you could simplify it with
`var fData = fd[el] || (sd && sd[el]);`
and remove the `else if` part

www/common/make-backup.js
```diff
@@ -286,26 +307,32 @@ define([
...
             var el = root[k];
+            let staticData;
             if (typeof el === "object" && el.metadata !== true) { // if folder
                 var fName = getUnique(sanitize(k), '', existingNames);
                 existingNames.push(fName.toLowerCase());
-                return void makeFolder(ctx, el, zip.folder(fName), fd);
+                return void makeFolder(ctx, el, zip.folder(fName), fd, sd);
             }
             if (ctx.data.sharedFolders[el]) { // if shared folder
+                staticData = ctx.sf[el].static;
```
yflory, line 322: `staticData is only used inside this `if`. You could declare it here with `let` directly.

www/common/make-backup.js
```diff
@@ -169,10 +169,18 @@ define([
...
             });
         }
-
-        var href = (fData.href && fData.href.indexOf('#') !== -1) ? fData.href : fData.roHref;
-        var parsed = Hash.parsePadUrl(href);
-        if (['pad', 'file'].indexOf(parsed.hashData.type) === -1) { return; }
+        var href;
+        var parsed;
+        var linkRegex = new RegExp("^(http|https)://");
+        if (fData.href && linkRegex.test(fData.href)) {
```
yflory, line 175: This *should* work but this isn't a very clean way to detect if this is a link. If someone created a pad at the very beginning of CryptPad and this pad's href wasn't properly migrated to the new format, it's possible that they still have a pad with an absolute URL in their drive `https://cryptpad.fr/pad/#...`.
It's also possible that in the future we'll want to have absolute URL for pads stored at a different place.

Another way to detect if the item is a link is to simply check if `channel` exists:
```javascript
if (!fData.channel) {
  // this is a link
} else {
```

yflory (changes requested):

yflory (approved):

PR #1697, opened 2024-10-29
Prevent links in trash from disappearing after drive reload

### LINKED ISSUE #1696: Links disappear from Trash
### Description

When a link is trashed and the browser page is refreshed, it then disappears from the trash (& Drive in general) entirely.

### Steps to reproduce

1. Create link in Drive
2. Move link to trash
3. Refresh page
4. Link is no longer anywhere in the Drive, including the trash

### Relevant logs

```shell
"An element in TRASH doesn't have associated data [xxx]"
"Your file system was corrupted. It has been cleaned so that the pads you visit can be stored safely."
userObject.js:160
```

### Expected behavior

When any Drive element, including links, is moved to the trash, it should persist there until it is removed/destroyed or the trash is emptied.

_No response_

_No response_

### Extensions

n/a

### Version

2024.9.0

zuzanna-maria:
* When a link is moved to the trash, and the browser page refreshed, it now stays in the trash until it is removed / destroyed instead of disappearing (fixes #1696)

### DIFF
--- a/www/common/outer/userObject.js
+++ b/www/common/outer/userObject.js
@@ -698,7 +698,7 @@ define([
                     }
                     if (exp.isFolder(obj.element)) { fixRoot(obj.element); }
                     if (typeof obj.element === "number") {
-                        var data = files[FILES_DATA][obj.element];
+                        var data = files[FILES_DATA][obj.element] || files[STATIC_DATA][obj.element];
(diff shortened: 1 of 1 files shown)


davidbenque:
Tested the behavior and seems OK but will leave someone to review the JavaScript

Issue #1694, opened 2024-10-24
Kanban items in a column are not fully scrollable on mobile browsers


tigexplorer:
### Description

If the amount of items (cards) in a column is more then a mobile screen can display, they are not fully and smooth scrollable. 

Swiping to scroll moves the cards a little bit and then holds. The UI seems to freeze. After a while one can swipe and scroll again a little bit.

Testet with Fennec/ Firefox on Android and Safari on iOS.

Here is a test kanban:
https://cryptpad.adminforge.de/kanban/#/2/kanban/view/TIopyJlzUyvEfPU-68XN6FiAMRo5E2iVWHkcWY+WMmQ/

While only one OS is selectable: It is the same for iOS

### Steps to reproduce

1. Fill a column with cards so that the list is larger then a mobile screen.
2. Try to swipe down or up

### Expected behavior

The card list should slide easily through the entire list when swiping 

### Which operating system are you using?

Google Android

### What browsers are you seeing the problem on?

Mozilla Firefox

### Extensions

no

### Version

2024.9.0

orogenic:
overall mobile UX is questionable IMO. long hold to enter a multi-select/move mode would make more sense IMO? multi-select is a separate feature wish, to open the door for bulk editing (assuming it doesn't exist).

PR #1705, opened 2024-11-03
kanban todo list no scroll in mobile fix

rohit-kumar-72:
fixed bug #1694


rohit-kumar-72:
here is a short recording.

https://github.com/user-attachments/assets/2fe32c7e-1062-424d-860f-84a5571baaf8

davidbenque:
Hi and thanks for looking into this issue. We are planning to address this in the next release cycle but with a more involved fix than just touch events, see #1727 
This will create a mode for touchscreens where you can move cards with buttons and use touch only to move/scroll around.

Issue #1606, opened 2024-08-26
Main iframe title does not help navigation


Sereza7:
### Description

Iframes should have text alternatives. This is a continuation to https://github.com/cryptpad/cryptpad/issues/1123. 
The main iframe's title is `iframe`, which does not help much with navigation. This solution might fulfill WCAG requirements, it does not improve the accessibility of this structure.

![Screenshot from 2024-08-26 17-31-57](https://github.com/user-attachments/assets/0283cdc0-9f88-42e8-a38d-ef5177017f20)

### Steps to reproduce

1. Create a document
2. Open the document
3. Check the programmatic name of the main iframe, either with a screen reader or with your browser inspector.

### Expected behavior

The main iframe has a proper name, such as `Main content` or `Main`.

### Which operating system are you using?

Linux/BSD/UNIX

### What browsers are you seeing the problem on?

Mozilla Firefox

### Extensions

Too many :')
Ublock and Consent-O-Matic are usually messing some features up but I doubt they'd be relevant here.

### Version

2024.6.1

PR #1706, opened 2024-11-04
Change iframe title

dariiing:
This PR updates the iframe title, closing #1606.
The change has been tested with NVDA.


davidbenque:
A couple of issues:
- "main content" or any other string used as title for the iframe would need to be translated (not hard-coded in English)
- I'm actually doubting whether we need a title for this iframe since it is always there and encloses _all of CryptPad_ for security reasons

I would suggest to try with an empty string as title attribute. This will comply with accessibility tests while not adding unnecessary content for screen reader users.

PR #1711, opened 2024-11-12
Form answers Drive download

### LINKED ISSUE #1710: Drive downloads should include responses to Form files
### Description

Copying over comment by @Asamartino from comment thread in #1051:

> I'm following the same steps to reproduce [downloading Drive contents as .zip] as @Mthld. It doesn't download the responses to forms -> I had to download each file manually.

### Steps to reproduce

1. Download / back up Drive contents as .zip file from Settings panel
2. If Drive includes Form files with responses, the .zip only includes the Form file itself (without responses). To download the responses for each Form, the user must do so manually from each individual Form file

### Relevant logs

_No response_

### Expected behavior

Copying over comment by @Asamartino from comment thread in #1051:

> All files should be exported successfully and present in the drive archive (including the responses to forms).

_No response_

_No response_

### Extensions

n/a

### Version

2024.9.0

zuzanna-maria:
- Fixes #1710 (Form responses now included in Drive backup .zips)


www/form/export.js
```diff
@@ -84,14 +73,14 @@ define([
...
+
                 order.forEach(function (key) {
                     if (!form[key]) { return; }
                     var type = form[key].type;
                     if (!TYPES[type]) { return; } // Ignore static types
                     var id = `q${i++}`;
-                    if (TYPES[type].exportCSV) {
-                        data[id] = TYPES[type].exportCSV(msg[key], form[key]);
-                        return;
+                    if (type === 'date') {
```
yflory, line 82: And what about all the other types that have special cases for their exports?

www/form/export.js
```diff
@@ -162,13 +150,55 @@ define([
...
+                var results = new Blob([arr], { type : "application/json" });
+                var opts = {
+                    binary: true,
+                };
+                zip.file(fileName, results, opts);
+            });
+        }
+
+        var json = Util.clone(content || {});
+            delete json.answers;
```
yflory, line 197: Indentation issue

www/form/export.js
```diff
@@ -162,13 +150,55 @@ define([
...
+        if (sframeChan && content.form) {
+            var _answers = content["answers"];
+            _answers['href'] = parsed.hash;
+            _answers['password'] = parsed.password;
+            _answers['drive'] = true;
+            var answers;
+            sframeChan.query("Q_FORM_FETCH_ANSWERS", _answers, function (err, obj) {
+                answers = obj && obj.results;
+                var rawName = content.metadata.title || 'File';
+                var fileName = getUnique(sanitize(rawName + ' (answers)'), '.json', existingNames);
```
yflory, line 184: Several issues here:
* After a quick test, I ended up with `File (answers)`, which means I didn't know which form it was about
* You're using hardcoded text so only English is supported. You should use translation keys!
* I think it's a bad idea to export 2 files for the same form. The default export is a way to make a backup of your form as a JSON file and be able to restore/re-import it later or in a new form. If we're doing a drive download, I think the "results" file with the questions and all the responses is enough. The drive download is not meant to be re-imported (and cannot be re-imported for most apps)

www/form/main.js
```diff
@@ -108,161 +104,7 @@ define([
...
-                        Nacl.util.decodeBase64(curvePrivate)
-                    );
-                    return channel === Nacl.util.encodeUTF8(u8_plain);
-                } catch (e) {
-                    console.error(e);
-                    return false;
-                }
-            };
-
             var deleteLines = false; // "false" to support old forms
```
yflory, line 107: I'm a bit scared about this "deleteLines" duplicated in sframe-common-outer.js and main.js. This value can be changed from 2 places which were both in this file, but now the other part is in sframe-common-outer.js.
I think there is a risk that, in main.js, we call some code that uses "deleteLines" before it is set to the correct value (because only the variable in sframe-common-outer would be updated).
This will have to be properly tested!

www/common/sframe-common-outer.js
```diff
@@ -283,6 +285,174 @@ define([
...
+                                Nacl.util.decodeBase64(pub),
+                                Nacl.util.decodeBase64(curvePrivate)
+                            );
+                            return channel === Nacl.util.encodeUTF8(u8_plain);
+                        } catch (e) {
+                            console.error(e);
+                            return false;
+                        }
+                    };
+                    var deleteLines = false; // "false" to support old forms
```
yflory, line 310: I think this `deleteLines` is a problem.
First "issue" is that in your case, the variable is initialized as `false` and then it is set to `true` when we encounter a form for which we can delete our response. Then it is never reset back to `false` for the other forms. It is a minor issue because I don't think it is used when we just want to export the responses and when you're in a form app, you only work with a single form.
The other issue, which a bit more important: you duplicated the `deleteLines` variable from form/main.js. In this PR, this variable is used in both files. But since you moved half of the code, each file is losing a way to set this variable to true. We may end up in cases where it is true in `main.js` and false here (or the opposite). I still it would be cleaner to move all the form command handler into their own file, and calling this file from sframe-common-outer.js and main.js to have it all in a single place.

PR #1720, opened 2024-11-22
Formatted tables and strikethrough text in Pad .md exports

### LINKED ISSUE #1531: Exporting tables to markdown does not work
### Description

I tried to export a document containing a table to markdown.  Unfortunately, the table is flattened in the markdown, like so:

```md
00  

01  

10  

11
```

In contrast, exporting to html preserves th table:

```html
<table style="width: 500px;" cellspacing="1" cellpadding="1" border="1"><tbody><tr><td>00<br type="_moz"></td><td>01<br type="_moz"></td></tr><tr><td>10<br type="_moz"></td><td>11<br type="_moz"></td></tr></tbody></table>
```

### Steps to reproduce

1. Create a rich text document
2. Insert a table
3. Export to markdown

### Relevant logs

_No response_

### Expected behavior

I'd like the table to be exported to markdown, like so:

```md
| 00 | 01 |
| 10 | 11 |
```

_No response_

_No response_

### Extensions

None.

### Version

2024.6.0

zuzanna-maria:
* Pad documents exported in .md format now include table and strikethrough text formatting (fixes #1531)

### DIFF
--- a/www/pad/export.js
+++ b/www/pad/export.js
@@ -118,7 +118,7 @@ define([
                                 var rowContent = Array.from(childNode.childNodes);
                                 var indexOf = Array.prototype.indexOf;
                                 var index = childNodeArr.length > 1 ? indexOf.call(node.childNodes, rowNode) : indexOf.call(rowNode.childNodes, childNode);
-                                var row = '';
+                                var row = '|';
                                 var rowLength = rowContent.filter(Boolean).length;
                                 for (var i =0; i < rowLength; i++) {
                                     var cell = rowContent[i];
@@ -137,7 +137,7 @@ define([
                                 var newRow = row.concat('\n');
                                 if (index === 0) {
                                     var separator = '|-';
-                                    newRow += `${separator.repeat(rowLength)}\n`;
+                                    newRow += `${separator.repeat(rowLength)}|\n`;
                                 }
                                 table += newRow;
                                 return newRow;


www/pad/export.js
```diff
@@ -96,7 +96,34 @@ define([
             if (ext === ".md") {
                 var md = Turndown({
                     headingStyle: 'atx'
-                }).turndown(toExport);
+                }).addRule('table', {
+                filter: ['tr'],
```
yflory, line 100: Small indentation issue starting here

www/pad/export.js
```diff
@@ -96,7 +96,34 @@ define([
...
+                replacement: function (content, node) {
+                    var indexOf = Array.prototype.indexOf;
+                    var index = indexOf.call(node.parentNode.childNodes, node);
+                    var rowContent = node.innerHTML.replace(/<td>/g, '').replace(/<br>/g, ' ').split('</td>');
```
yflory, line 104: 1. Removing the `<br>` here will remove new lines inside cells. This is acceptable since currently exports are not working but it would be nice to be able to fix this.
2. This won't work if the table contains "header" cells which are `<th>` instead of `<td>`
3. I think it would be easier to work with the DOM elements directly instead of treating the HTML as a string

yflory (changes requested):

yflory (approved):

davidbenque:
The markdown tables seem to be missing a pipe character at the start of rows that have content, and at the end of the empty row after the header. 

The CryptPad Code app seems to render the table anyways, but other apps might not (I tested Typora which didn't work). Github also seems to be fine with the example below, so maybe this is a niche issue with the apps I use and my way of writing tables.

original Rich Text table:  
image

Markdown export:  
```
test in|value **number** (kg)|Date|
|-|-|-
_yes_|<u>this is a weird one</u>|2024|
no|old option is ~not valid~ anymore|2025|
```

Rendered on GitHub:
test in|value **number** (kg)|Date|
|-|-|-
_yes_|<u>this is a weird one</u>|2024|
no|old option is ~not valid~ anymore|2025|

PR #1721, opened 2024-11-22
Calendar reminders in notification panel

### LINKED ISSUE #1374: Calendar notifications do not appear in the notification panel
### Description

When a user receives a notification about a calendar event, then uses the notification button to access the notification panel, the notification is still present in the toolbar drop-down but not in the panel.

### Steps to reproduce

1. Receive a notification about a Calendar event (past or present) and view it using the toolbar notiification button dropdown.
2. Access the notification panel. 
3. The notification is not visible in the panel.

### Relevant logs

_No response_

### Expected behavior

The Calendar notifications which a user receives should be visible in the notification panel until dismissed.

_No response_

_No response_

### Extensions

n/a

### Version

5.6.0

zuzanna-maria:
Calendar reminders are now included in the notification panel (fixes #1374)

### DIFF
--- a/www/notifications/inner.js
+++ b/www/notifications/inner.js
@@ -145,7 +145,7 @@ define([
 
         common.mailbox.subscribe(["notifications", "reminders"], {
             onMessage: function (data, el) {
-                el.innerHTML = el.innerHTML.replace('<i', '<span class="cp-avatar"><span class="cp-avatar-default" alt="" aria-hidden="true"><i').replace('/i>', '/i></span></span>')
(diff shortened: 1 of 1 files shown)


www/notifications/inner.js
```diff
@@ -143,8 +143,9 @@ define([
             $(loadmore).click();
         }
 
-        common.mailbox.subscribe(["notifications"], {
+        common.mailbox.subscribe(["notifications", "reminders"], {
             onMessage: function (data, el) {
+                el.innerHTML = el.innerHTML.replace('<i', '<span class="cp-avatar"><span class="cp-avatar-default" alt="" aria-hidden="true"><i').replace('/i>', '/i></span></span>');
```
Chouhartem, line 148: Maybe better to use some DOM manipulation + hyperscript here to have a more robust/readable solution than `sed`ing the `innerHTML` directly?

Chouhartem (approved):

Issue #625, opened 2020-10-21
NFR - ability to disable anonymous access on self hosted instance


reetp:
**Describe the bug**
Allowing anonymous access on self hosted could have potential legal implications for the hoster.

e.g. it could be used by nefarious users to paste indecent or illegal material.

Particularly serious for any companies hosting this.

**Where did it happen?**
Any self hosted instance.

**To Reproduce**
Anonymously create any form of pad. Give a link to your your partner in crime.

Paste something illegal for them.

**Expected behavior**
Simple method to disable anonymous access

ohader:
seems like https://docs.cryptpad.fr/en/admin_guide/customization.html#restricting-guest-access, by extending `customize/application_config.js` with the following assignment solved for me: 

```js
AppConfig.registeredOnlyTypes = AppConfig.availablePadTypes;
```

ohader:
... the assignment above also disables the possibility to share read-only documents to "the public". Did not find anything else for disabling anonymous write-access, but keeping the ability to still read shared content.

Wolf480pl:
> AppConfig.registeredOnlyTypes = AppConfig.availablePadTypes;

doesn't this being in `customize/application_config.js` mean it's only client-side?

Wolf480pl:
Looks like this would be a good place to prevent anonymous users from creating new pads:
https://github.com/cryptpad/cryptpad/blob/main/lib/hk-util.js#L741

Enteee:
This is probably a duplicate of #704  and could be closed maybe?

reetp:
> This is probably a duplicate of #704  and could be closed maybe? 
> ... or 704 could be closed :)

Ahhhhh indeed. Damn how did I miss that?

Yes, one or the other.

The essential point here is it should be an easy one click disable on the server side for an admin.

Sad as we can't use it as it is.

Issue #704, opened 2021-03-16
Is there any way to prevent anonymous users from creating pads but retain their ability to view and edit existing ones via link?


vocux:
I have just set up my own self hosted instance (awesome so far!) and like some other users I seek to limit anonymous access in order to prevent abuse.

As I understand it, you can already:

- disable almost any pad / functionality for anonymous users via custom application_config.js (using registeredOnlyTypes)
- disable anonymous storage (again custom application_config.js), preventing anonymous users from uploading and retaining pads,
- prevent users from registering, at least with [this](https://github.com/xwiki-labs/cryptpad/issues/242#issuecomment-572370574) nginx workaround.

So you can already pretty much lock down your instance from anonymous access, but can you prevent anonymous users from **creating** pads while retaining their ability to view (or edit) existing pads (depending on what kind of link it is)? This would be similar to how for example Nextcloud works in this regard, with private files that can be shared via a public link. Using registeredOnlyTypes, anonymous clients cannot view anything anymore.

Is this possible somehow? Otherwise I'd kindly request this feature.

Best wishes!

Enteee:
Would "disable anonymous storage" (point 2) not be sufficient for this?

vocux:
> 
> 
> Would "disable anonymous storage" (point 2) not be sufficient for this?

After a quick test I found that you can still create pads as an anonymous user and share them via link, even with the disabled storage setting turned on. So no, this would not be sufficient.
I guess those pads will be deleted after a number of days, this set in the "inactiveTime" variable, right? But you can't turn this setting to 0 either, as this means that expiration is turned off completely?

Enteee:
> After a quick test I found that you can still create pads as an anonymous user and share them via link, 

Thank you for testing this. It's a bummer that this does not work. It's a bit of a show-stopper for self hosted cryptpad instances. I think the share functionality is really cool,  but at the same time I also don't random internet people to be able to create pads and fill my backup or even worse, share illegal stuff via my servers.

ansuz:
[this commit](https://github.com/xwiki-labs/cryptpad/commit/f5c029e937de0bf02079e2f1e7ebca2bb09f7558) makes it possible to disable the creation of pads by anonymous users by adding `AppConfig.disableAnonymousPadCreation = false;` to `customize/application_config.js`.

This check is only on the client, so it won't stop determined adversaries, but it's probably good enough for most purposes.

Enteee:
[...]

PR #1724, opened 2024-12-01
Add option to disable anonymous pad creation on the server side

Wolf480pl:
This is my first attempt at fixing #625 

Looks like concept of "user" is very ambiguous in cryptpad, so I assumed `Users.getAll` are the registered users.
As far as I understand, in registration-disabled mode, these include all users created by accepting an invitation.
The admin can also add users manually to that list, which sounds useful.
Not sure how this handles SSO though.
Also, should I allow admins to create pads, even if they're not in Users?

I'd appreciate suggestions about better ways of obtaining something akin to a list of registered users.

Also, AFAIU this only prevents creation of channels (which seems to include userobjects, drive, etc) but not pins, blocks, and possibly other kinds of things that could be abused by an attacker to store arbitrary data.
I don't have a good grasp of those, or a good way to test for that (looks like the official client won't try to do these things when not logged in) so I'd also appreciate any pointers on how to handle those things.

Also, I don't have much experience with JS, so my code probably looks very awkward and unidiomatic - sorry about that.

### DIFF
--- a/config/config.example.js
+++ b/config/config.example.js
@@ -174,6 +174,17 @@ module.exports = {
     ],
 
     /* =====================
+     *      PERMISSIONS
+     * ===================== */
+
+    /*  Prevent anonymous users from creating new pads (including user objects,
+     *  and other internal pads).
+     *  This setting works on the server side, so it should in theory
+     *  work against malicious clients.
+     */
+    // disableAnonymousPadCreationServerside: false,
+
+    /* =====================
      *        STORAGE
      * ===================== */
 
--- a/lib/env.js
+++ b/lib/env.js
@@ -119,6 +119,7 @@ module.exports.create = function (config) {
         httpAddress: typeof(config.httpAddress) === 'string'? config.httpAddress: 'localhost',
         websocketPath: config.externalWebsocketURL,
         logIP: config.logIP,
+        disableAnonymousPadCreationServerside: config.disableAnonymousPadCreationServerside,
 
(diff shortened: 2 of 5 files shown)


Wolf480pl:
Also it breaks invites, I'll try to fix that.

Wolf480pl:
I just realized it also breaks login, because edPrivate is not stored in the block

Enteee:
@Wolf480pl  this would be such a cool feature !
See: #704 which is related to this.

Wolf480pl:
I was trying to wrap my head around the frontend code responsible for the login and registration, and how it stores account data on the server, in order to make invites work with this change. But it's quite complex, and I ran out of time. I may try to return to this in the future, but if anyone else wants to give it a try that'd be cool too.

f01c33:
Looks like your PR blocks not only registration, but any file creation, i can't see a way for the server to validate that a specific user is actually registered, we'd have to have pubkey in memory? how? is that even private info? to be honest trying to fix this made me lose all hope of using this software on the open internet, how could we possibly stop abuse with its current arch?

Wolf480pl:
Honestly, I didn't fully understand how cryptpad works when I was writing this, and what little I did understand, I forgot by now.
Feel free to treat this as abandoned.

f01c33:
Not your fault, I'm just baffled that they though it was a good idea to
make it a secret even the pub keys, took anonymity as gospel even if that
means not having authorization.

Issue #1409, opened 2024-02-08
UI on mobile: Instead of scrolling a list, it drags and moves the card


ldeluigi:
I feel this deserves a standalone issue.

_Originally posted by @ldeluigi in https://github.com/cryptpad/cryptpad/issues/1372#issuecomment-1934254903_

Issue #1372, opened 2024-01-08
Kanban lists difficult to drag on mobile 


zuzanna-maria:
### Description

Difficult to drag Kanban lists - when attempting to drag a Kanban list from one side of the board to the other, can't do it in one movement (screen doesn't automatically scroll to the side) - have to drop and re-drag the board several times

### Steps to reproduce

1. Access Kanban on mobile
2. Attempt to drag the list on the leftmost side of the board to the rightmost side
3. Since the board doesn't automatically scroll to the side when attempting the drag, it is necessary to drop the list along the way and repeat until the desired position is reached

### Expected behavior

Kanban lists should be smoothly draggable from one side of the board to the other, without having to drop them along the way.

Alternately, 

> having two small buttons on the top right of every items ◀️▶️ (lists and tiles).

### Extensions

n/a

### Version

5.6.0

ldeluigi:
Same for vertical scrolling! Instead of scrolling under a list, it drags and moves the card!

**It should wait at least a second before considering the swipe a drag instead of a scroll.**

PR #1727, opened 2024-12-03
Buttons for moving Kanban boards and items on mobile (WIP)

zuzanna-maria:
- Fixes issues related to moving Kanban boards and items on touchscreen devices by adding arrow buttons and disabling drag (#1694, #1372, #1409)

### DIFF
--- a/www/kanban/inner.js
+++ b/www/kanban/inner.js
@@ -559,6 +559,71 @@ define([
                 e.stopPropagation();
             }).appendTo($(el).find('.kanban-board-header'));
         });
+        $container.find('.kanban-board').each(function (i, el) {
+            $('<button>', {
+                'class': 'kanban-edit-item fa fa-arrow-left',
+                'alt': '', 
+            }).click(function (e) {
+                var boardList = $container[0].childNodes[0].childNodes[0].childNodes
+                var newKanban = boardList[boardList.length-1]
+                var boardsArray = Array.from(boardList)
+                boardsArray.slice(boardsArray.length-1, 1)
+                var moveBoards = function (arr, oldIndex, newIndex) {
+                    if (newIndex >= arr.length) {
+                        var k = newIndex - arr.length + 1;
+                        while (k--) {
+                            arr.push(undefined);
+                        }
+                    }
+                    arr.splice(newIndex, 0, arr.splice(oldIndex, 1)[0]);
+                    return arr;
+                };
+                if (i === 0) {
+                    moveBoards(boardsArray, i, boardsArray.length-1)
+                } else {
+                    moveBoards(boardsArray, i, i-1)
+                }
+                var index = boardsArray.indexOf(el)
+                i = index
+                var boardContainer = $container[0].childNodes[0].childNodes[0]
+                boardContainer.innerHTML = ''; 
+                boardsArray.forEach(board => {
+                    boardContainer.appendChild(board);
+                });
+                boardContainer.appendChild(newKanban);                
+            }).appendTo($(el).find('.kanban-board-header'));
+        });
+         $container.find('.kanban-board').each(function (i, el) {
+            $('<button>', {
+                'class': 'kanban-edit-item fa fa-arrow-right',
+                'alt': '', 
+            }).click(function (e) {
+                var boardList = $container[0].childNodes[0].childNodes[0].childNodes
+                var newKanban = boardList[boardList.length-1]
+                var boardsArray = Array.from(boardList)
+                boardsArray.slice(boardsArray.length-1, 1)
+                var moveBoards = function (arr, oldIndex, newIndex) {
+                    var element = arr[oldIndex];
+                    arr.splice(oldIndex, 1);
+                    arr.splice(newIndex, 0, element);
+                };
+                if (i === boardsArray.length-2) {
+                    moveBoards(boardsArray, i, 0)
+                } else {
+                    moveBoards(boardsArray, i, i+1)
+                }
+                var boardContainer = $container[0].childNodes[0].childNodes[0]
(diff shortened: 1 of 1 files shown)


dariiing:
Currently, scrolling through the boards is only possible on mobile devices. It should be adjusted to have the same behavior across all devices. Similarly, dragging boards and items should also be consistently available on all devices.  The new arrow functionality can be a solution, but I assumed  it's primarily intended for keyboard users.

dariiing (changes requested):
Currently, scrolling through the boards is only possible on mobile devices. It should be adjusted to have the same behavior across all devices. Similarly, dragging boards and items should also be consistently available on all devices. The new arrow functionality can be a solution, but I assumed it's primarily intended for keyboard users.

www/kanban/app-kanban.less
```diff
@@ -170,25 +170,45 @@
...
+        cursor: pointer;
         &:focus-visible {
             outline: @variables_focus_style;
         }
-        @media (hover: none) {
-            margin-right: 20px;
-        }
         @media(max-width:500px) or (max-height:700px)
         {
             margin-right:0px;
```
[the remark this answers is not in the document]
zuzanna-maria: This code was removed to get rid of the awkward sizing of items on smaller screens

www/kanban/inner.js
```diff
@@ -534,6 +536,137 @@ define([
         UI.openCustomModal(editModal.modal);
     };
 
+    addMoveElementButton = function (framework, kanban) {
+        var $container = $(kanban.element);
+        var drag = $('#toggle-drag-off').attr('class').indexOf('toggle-active') !== -1 ? false : true;
+        kanban.options.dragBoards = drag;
+        kanban.options.dragItems = drag;
+        if (drag === false) {
+            if (!kanban) { return; }
```
yflory, line 545: If kanban is falsy, this code will throw an error 5 lines above with `var $container = $(kanban.element);`

www/kanban/inner.js
```diff
@@ -534,6 +536,137 @@ define([
...
+
+            var move = function (arr, oldIndex, newIndex) {
+                arr.splice(newIndex, 0, arr.splice(oldIndex, 1)[0]);
+                _updateBoards(framework, kanban, kanban.options.boards, false);
+            };
+
+            var moveBetweenBoards = function (nextBoardItems, elId, boardItems, index, boards, boardId) {
+                nextBoardItems.unshift(elId);
+                boardItems.splice(index, 1);
+                boards.data[boardId].item = boardItems;
```
yflory, line 556: This won't do anything and the line can be removed. `boardItems` is only a reference to the same object that was already available from `boards.data[boardId].item`.

www/kanban/inner.js
```diff
@@ -534,6 +536,137 @@ define([
...
+                _updateBoards(framework, kanban, boards, false);
+                $(`.kanban-item[data-eid="${elId}"]`)[0].scrollIntoView();
+            };
+
+            var shiftItem = function (direction, el) {
+                var board = el.parentNode.parentNode.parentNode;
+                var boards = kanban.options.boards;
+                var elId = parseInt($(el).attr("data-eid"));
+                var boardId = parseInt($(board).attr("data-id"));
+                var boardItems = direction === ('up' || 'down') ? kanban.options.boards.data[boardId].item : boards.data[boardId].item;
```
yflory, line 566: This very complex line is the same as: `var boardItems = boards.data[boardId].item;`.

`direction === ('up' || 'down')` this is valid JavaScript but it doesn't really makes sense: `('up' || 'down')` is just `'up'`
And `kanban.options.boards` is exactly the same as `boards` (defined 3 lines above)

www/kanban/inner.js
```diff
@@ -534,6 +536,137 @@ define([
...
+            var moveBetweenBoards = function (nextBoardItems, elId, boardItems, index, boards, boardId) {
+                nextBoardItems.unshift(elId);
+                boardItems.splice(index, 1);
+                boards.data[boardId].item = boardItems;
+                _updateBoards(framework, kanban, boards, false);
+                $(`.kanban-item[data-eid="${elId}"]`)[0].scrollIntoView();
+            };
+
+            var shiftItem = function (direction, el) {
+                var board = el.parentNode.parentNode.parentNode;
```
yflory, line 562: :open_mouth: 
Maybe search the element by class? or use `$(el).closest('.cp-class-name')` to search a class only amongst the ancestors.
Using this chain of parentNode makes it unreliable in case we update the structure.

yflory (changes requested):
In addition to the comments here, most of the requested changes from the "form" PR #1753 also apply here (prefix the ID, deduplicate code, avoid checking the state using the class of a button)

yflory (approved):

dariiing:
While testing:
- the edit board title/ edit item buttons don't work as intended when navigating via keyboard (they are focusable, but when I press Enter nothing happens, it should mimic clicking)
- the newly added buttons ( Move left/right/up down) also don't work as previously mentioned
- the tooltip of some buttons is undefined
- the buttons introduced this release are not described/announced correctly when using screen readers
- the keyboard navigation doesn't follow the visual order for the newly added buttons

https://github.com/user-attachments/assets/d9d06614-669a-4417-8c3b-7ca5932f4294

davidbenque:
Applied some fixes to this branch

-[x] Corrected markup from `h('.button' ...` to `h('button' ...` to actually create buttons instead of `<div class="button"` this restores a lot of the default accessibility 
- [x] Moved icons to their own `i` element and applied `aria-hidden` only to that 
- [x] Corrected some typos in translation key names that caused them to be undefined

dariiing (approved):

PR #1739, opened 2024-12-14
Sort files and folders with "natural" sort

jfly:
With this change, we now sort "folder 2" *before* "folder 10".


jfly:
@mathilde-cryptpad, rebased!

PR #1740, opened 2024-12-16
Add `--check`, `--rdfind`, `--no-rdfind` options to `install-onlyoffice.sh`

jfly:
`--check` is useful for anyone who wants to check if their OnlyOffice
installation is up to date without actually installing it.

The `--rdfind/--no-rdfind` commands serve a few purposes:

1. It makes it clear that `rdfind` can still run even if you specify `--check`.
2. People can ensure that `rdfind` runs by specifying `--rdfind`, which
   helps if they accidentally uninstall `rdfind`.
3. People can now skip running `rdfind` even if they have `rdfind`
   installed.

### DIFF
--- a/install-onlyoffice.sh
+++ b/install-onlyoffice.sh
@@ -37,7 +37,17 @@ main() {
     install_x2t v7.3+1 ab0c05b0e4c81071acea83f0c6a8e75f5870c360ec4abc4af09105dd9b52264af9711ec0b7020e87095193ac9b6e20305e446f2321a541f743626a598e5318c1
 
     rm -rf "$BUILDS_DIR"
-    if command -v rdfind &>/dev/null; then
+
+    if [ "${RDFIND+x}" != "x" ]; then
+        if command -v rdfind &>/dev/null; then
+            RDFIND="1"
+        else
+            RDFIND="0"
+        fi
+    fi
+
+    if [ "$RDFIND" = "1" ]; then
+        ensure_command_available rdfind
         rdfind -makehardlinks true -makeresultsfile false $OO_DIR/v*
     fi
 }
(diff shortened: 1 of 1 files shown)


install-onlyoffice.sh
```diff
@@ -166,22 +184,31 @@ install_version() {
 }
 
 install_x2t() {
-    ensure_command_available curl
-    ensure_command_available sha512sum
-    ensure_command_available unzip
```
jfly, line 171: I moved these checks down to right before we use the relevant tools. I'd argue this is better style (keep related things closer to gether). This is also helpful so the nix build (which installs OnlyOffice in its own way) doesn't have to depend on curl. (see https://github.com/NixOS/nixpkgs/pull/365514)

install-onlyoffice.sh
```diff
@@ -143,7 +174,17 @@ install_version() {
     local LAST_DIR
     LAST_DIR=$(pwd)
 
-    if [ ! -e "$FULL_DIR"/.commit ] || [ "$(cat "$FULL_DIR"/.commit)" != "$COMMIT" ]; then
+    local ACTUAL_COMMIT
+    if [ -e "$FULL_DIR"/.commit ]; then
+        ACTUAL_COMMIT="$(cat "$FULL_DIR"/.commit)"
+    fi
+
+    if [ "$ACTUAL_COMMIT" != "$COMMIT" ]; then
```
wginolas, line 182: When no OnlyOffice is installed, the script fails here, because `ACTUAL_COMMIT` is unset.

wginolas (changes requested):
Thanks for the improvements! I found a problem that I commented in the code.

jfly:
Sure thing, @mathilde-cryptpad! I've rebased this branch.

wginolas (approved):
Looks good to me now. Thank you!

Issue #1090, opened 2023-07-05
Profile page layout is misaligned


zuzanna-maria:
### Description
Profile page layout is misaligned - 

- The right side of the page is cut off

![Image](https://github.com/cryptpad/cryptpad/assets/34193395/59855254-b33b-4718-9a36-46f5c4f43636)

- When scrolling sideways, the background changes colour

![Image](https://github.com/cryptpad/cryptpad/assets/34193395/141b0406-1731-4701-9093-8da6d0d5a8f7)

### Steps to reproduce
1. Access 'Profile' section
2. (Scroll sideways)

### Relevant logs
n/a

### Expected behavior
All profile information should be included in the frame, or if scrolling sideways is necessary, background colour should be consistent

### Extensions
n/a

### Version
5.3.0

depperm:
I'm pretty sure this is because of https://github.com/cryptpad/cryptpad/blob/d9bc9ec3aae6848363f8cad98362685f4243ebc1/customize.dist/src/less2/include/sidebar-layout.less#L169

PR #1742, opened 2024-12-16
Fix profile layout on small screens

DianaXWiki:
This PR aims to make the profile layout responsive to smaller screen sizes, fixing [#1090 ](https://github.com/cryptpad/cryptpad/issues/1090)

### DIFF
--- a/www/common/translations/messages.json
+++ b/www/common/translations/messages.json
@@ -212,6 +212,7 @@
     "canvas_imageEmbed": "Embed an image from your computer",
     "profileButton": "Profile",
     "profile_upload": " Upload a new avatar",
+    "profile_remove_avatar": "Remove this avatar",
     "profile_uploadSizeError": "Error: your avatar must be smaller than {0}",
     "profile_uploadTypeError": "Error: your avatar type is not allowed. Allowed types are: {0}",
     "profile_error": "Error while creating your profile: {0}",
--- a/www/common/translations/messages.json
+++ b/www/common/translations/messages.json
@@ -212,7 +212,6 @@
     "canvas_imageEmbed": "Embed an image from your computer",
     "profileButton": "Profile",
     "profile_upload": " Upload a new avatar",
-    "profile_remove_avatar": "Remove this avatar",
     "profile_uploadSizeError": "Error: your avatar must be smaller than {0}",
     "profile_uploadTypeError": "Error: your avatar type is not allowed. Allowed types are: {0}",
     "profile_error": "Error while creating your profile: {0}",
--- a/www/profile/app-profile.less
+++ b/www/profile/app-profile.less
@@ -6,6 +6,7 @@
 
 @import (reference) '../../customize/src/less2/include/framework.less';
 @import (reference) '../../customize/src/less2/include/sidebar-layout.less';
+@import (reference) '../../customize/src/less2/include/variables.less';
 
 &.cp-app-profile {
 
@@ -115,8 +116,7 @@
         }
         .cp-app-profile-link {
             font-size: 25px;
-            vertical-align: middle;
-            display: inline-block;
+            display: inline;
             overflow-wrap: anywhere;
         }
         .cp-app-profile-displayname, .cp-app-profile-link {
@@ -136,6 +136,11 @@
         & > button:empty {
             margin-left: 25px !important;
         }
+        @media screen and (max-width: @browser_media-medium-screen) {
+            .cp-app-profile-link-code {
+                min-width: 0;
+            }
+        }    
     }
 
     div.cp-app-profile-mute-container.cp-app-profile-section {
--- a/www/profile/app-profile.less
+++ b/www/profile/app-profile.less
@@ -24,6 +24,7 @@
     #cp-app-profile-header {
         display: flex;
         flex-flow: row wrap;
+        justify-content: center;
         #cp-app-profile-rightside {
             flex: 1;
             display: flex;
@@ -32,17 +33,15 @@
         }
     }
     #cp-app-profile-avatar {
-        width: 300px;
-        //height: 350px;
         margin: 10px;
         margin-right: 20px;
         text-align: center;
+        max-width: 300px;
+        max-height: 300px;
+        margin-bottom: 5rem;
         &> span {
             display: inline-block;
             text-align: center;
-            height: 300px;
-            width: 300px;
-            border: 1px solid @cp_profile-border;
             border-radius: 4px;
             overflow: hidden;
             position: relative;
@@ -59,6 +58,7 @@
             max-width: 100%;
             max-height: 100%;
             vertical-align: top;
+            border-radius: @variables_radius_L;
         }
         media-tag {
             height: 100%;
@@ -81,11 +81,6 @@
         }
     }
 
-    @media screen and (max-width: 550px) {
-        #cp-app-profile-avatar {
-            margin: auto;
-        }
-    }
     .cp-app-profile-resizer {
         text-align: center;
     }
--- a/www/profile/app-profile.less
+++ b/www/profile/app-profile.less
@@ -39,7 +39,7 @@
         text-align: center;
         max-width: 300px;
         max-height: 300px;
-        margin-bottom: 1rem;
+        margin-bottom: 3.5rem;
         &> span {
             display: inline-block;
             text-align: center;
--- a/www/profile/app-profile.less
+++ b/www/profile/app-profile.less
@@ -90,6 +90,9 @@
     .cp-app-profile-displayname {
         margin-top: 1rem;
     }
+    .cp-app-profile-section:empty {
+        display: none;
+    }
     .cp-app-profile-displayname, .cp-app-profile-link, .cp-app-profile-section {
         margin-bottom: 1rem;
         input {
--- a/www/profile/app-profile.less
+++ b/www/profile/app-profile.less
@@ -89,11 +89,7 @@
     .cp-app-profile-displayname {
         margin-top: 1rem;
     }
-    div.cp-app-profile-link.hidden{
-        display:none;
-    }
-
-    .cp-app-profile-displayname, .cp-app-profile-link {
+    .cp-app-profile-displayname, .cp-app-profile-link, .cp-app-profile-section {
         margin-bottom: 1rem;
         input {
             width: 100%;
@@ -120,6 +116,8 @@
         .cp-app-profile-link {
             font-size: 25px;
             vertical-align: middle;
+            display: inline-block;
+            overflow-wrap: anywhere;
         }
         .cp-app-profile-displayname, .cp-app-profile-link {
             line-height: 40px;
@@ -140,14 +138,12 @@
         }
     }
 
-    div.cp-app-profile-link {
-        &.cp-app-profile-description-rendered,
-        &.cp-app-profile-mute-container {
-          &:has(p) {
+    div.cp-app-profile-mute-container.cp-app-profile-section {
+        p {
             margin-bottom: 0;
-          }
         }
-      }
+    
+    }
     .cp-app-profile-friend {
         display: flex;
         align-items: center;
@@ -174,6 +170,7 @@
         }
         #cp-app-profile-description-info {
             overflow-wrap: anywhere;
+            margin-top: 1rem;
         }
         .cp-app-profile-description-ok, .cp-app-profile-description-spin {
             position: absolute;
--- a/www/profile/app-profile.less
+++ b/www/profile/app-profile.less
@@ -187,7 +187,7 @@
             font-family: monospace;
             font-size: 16px;
             line-height: initial;
-            width: 31rem;
+            max-width: 31rem;
             pre {
                 margin: 0;
                 font-family: inherit;
--- a/www/profile/app-profile.less
+++ b/www/profile/app-profile.less
@@ -168,9 +168,12 @@
     #cp-app-profile-description {
         position: relative;
         font-size: 16px;
-        margin-bottom: 20px;
+        //margin-bottom: 20px;
         .cp-app-profile-description-code {
             display: none;
+        }
+        #cp-app-profile-description-info {
+            overflow-wrap: anywhere;
         }
         .cp-app-profile-description-ok, .cp-app-profile-description-spin {
             position: absolute;
--- a/www/profile/app-profile.less
+++ b/www/profile/app-profile.less
@@ -23,6 +23,7 @@
 
     #cp-app-profile-header {
         display: flex;
+        flex-flow: row wrap;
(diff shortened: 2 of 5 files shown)


www/common/translations/messages.json
davidbenque, line 1: Please do not modify this file directly as this is handled by Weblate. Instead place new translation strings in `customize.dist/translations/messages.js` and make sure to add an `// XXX` comment on each line.

www/profile/app-profile.less
```diff
@@ -21,8 +21,14 @@
         display: none !important;
     }
 
+    #cp-app-profile-rightside:has(.cp-app-profile-friend-container) #cp-app-profile-link {
```
davidbenque, line 24: Please add a comment explaining why this needs to be hidden (element was already not visible but adding extra margin to the layout)

www/profile/app-profile.less
```diff
@@ -79,11 +84,15 @@
...
+
     .cp-app-profile-resizer {
         text-align: center;
     }
+    #cp-app-profile-displayname {
+        margin-top: 1rem;
+    }
     #cp-app-profile-displayname, #cp-app-profile-link {
-        margin: 10px 0;
+        margin-bottom: 1rem;
```
davidbenque, line 95: This is the expected margin bottom between buttons but other buttons are created and styled differently. You should apply the same margin-bottom to all buttons in that column. Watch out for classes that would affect all sidebar layout applications

www/profile/app-profile.less
```diff
@@ -142,17 +151,22 @@
...
+
     // I tried using flexbox but messed with how the pencil icon was displayed
     #cp-app-profile-invite-button {
         float: right;
     }
     .cp-app-profile-viewprofile-button {
-        margin-bottom: 5px !important;
+        margin-bottom: 1rem!important;
     }
     #cp-app-profile-description {
```
davidbenque, line 165: Please remove border and padding on `.cp-app-profile-description-rendered`

www/profile/inner.js
```diff
@@ -437,7 +437,7 @@ define([
         if (APP.readOnly) { return; }
 
         var button = h('button.btn.btn-primary', [
-            h('i.fa.fa-pencil'),
+            h('i', { 'class': 'fa fa-pencil', 'aria-hidden': 'true', 'aria-labelledby': Messages.profile_addDescription}),
             h('span', Messages.profile_addDescription)
         ]);
         APP.$descriptionEdit = $(button);
```
davidbenque, line 443: ln 452: Please move the text `p.cp-app-profile-info` to between the "view my profile" and "share" buttons as it seems more relevant

www/profile/app-profile.less
```diff
@@ -31,33 +37,32 @@
...
     }
     #cp-app-profile-avatar {
-        width: 300px;
-        //height: 350px;
         margin: 10px;
         margin-right: 20px;
         text-align: center;
+        max-width: 300px;
+        max-height: 300px;
+        margin-bottom: 5rem;
```
davidbenque, line 45: margin-bottom should be 1rem here

davidbenque:
In the profile editing view there is now an issue with overlapping username and button

image

davidbenque (changes requested):
- There is an overlap issue between username and button in the edit view

image

- Re-requested an accessibility review from @dariiing because there are remaining issues with button icons not being hidden from screen readers

dariiing (commented):
Icons have `aria-hidden`, so there isn't any issue there. However there are other problems that were not introduced with this PR:

1) Some toolbar elements in the description text box disappear on smaller screens: they are there, but not visually available. 
image

image

If I were to use the keyboard to navigate on a smaller screen, you can see that the elements are focusable, but hidden:

https://github.com/user-attachments/assets/01a1c2e4-1fb6-4c9a-8b83-3896b515d66b

2) The text area acts as a keyboard trap; we should implement the Esc key functionality as we have in other text areas.

3) The text area breaks or expands unexpectedly while typing.

https://github.com/user-attachments/assets/1bc0d5eb-2dad-402b-a233-f73daa34e2ae

www/profile/inner.js
```diff
@@ -101,40 +101,57 @@ define([
...
+        // Always create the description container
+        var $blockDescription = $('<div>', {id: DESCRIPTION_ID}).appendTo($container);
+        APP.$descriptionContainer = $blockDescription;
+        APP.$description = $('<div>', {
+            'id': 'cp-app-profile-description-info'
+        }).appendTo($blockDescription);
+        
+        // Add the share button only for editable profiles
+        if (!APP.readOnly) {
+            var $blockShare = $('<div>', {class: LINK_ID}).appendTo($container);
```
yflory, line 130: Same as before, LINK_ID doesn't really make sense here

www/profile/inner.js
```diff
@@ -101,40 +101,57 @@ define([
...
+            $('<button>', {
+                'class': 'btn btn-primary ' + VIEW_PROFILE_BUTTON, 'aria-labelledby': 'cp-share-button'
+            }).append(
+                h('i', { 'class': 'fa fa-share-alt', 'aria-hidden': 'true'})
+            ).append(
+                h('span', {'id': 'cp-share-button'}, Messages.shareButton)
+            ).click(function () {
```
yflory, line 137: Using hyperscript would make it easier to write and faster to render

www/profile/inner.js
```diff
@@ -101,40 +101,57 @@ define([
...
+            $button.click(function () {
+                window.open(url, '_blank');
             });
-        }).appendTo($container);
+        }
+        
+        // Always create the description container
+        var $blockDescription = $('<div>', {id: DESCRIPTION_ID}).appendTo($container);
+        APP.$descriptionContainer = $blockDescription;
+        APP.$description = $('<div>', {
```
yflory, line 124: Please try to use hyperscript when you add or rewrite code and you don't need to jQuery tools. It's faster to render, especially on mobile.

www/profile/inner.js
```diff
@@ -101,40 +101,57 @@ define([
...
+                h('span', {'id': 'cp-share-button'}, Messages.shareButton)
+            ).click(function () {
+                Clipboard.copy(url, (err) => {
+                    if (!err) { UI.log(Messages.shareSuccess); }
+                });
+            }).appendTo($blockShare);
+        }
     };
-
+
```
yflory, line 144: Spaces to remove here

www/profile/inner.js
```diff
@@ -101,40 +101,57 @@ define([
...
+                    if (!err) { UI.log(Messages.shareSuccess); }
+                });
+            }).appendTo($blockShare);
+        }
     };
-
+    
     var addDisplayName = function ($container) {
-        var $block = $('<div>', {id: DISPLAYNAME_ID}).appendTo($container);
+        var $block = $('<div>', {class: DISPLAYNAME_ID}).appendTo($container);
```
yflory, line 146: I don't understand this change. This makes an element and its child having the same class and applying styles to each other. I see that the class adds margins so each margin is applied twice.

www/profile/app-profile.less
```diff
@@ -79,11 +82,19 @@
...
+
     .cp-app-profile-resizer {
         text-align: center;
     }
-    #cp-app-profile-displayname, #cp-app-profile-link {
-        margin: 10px 0;
+    .cp-app-profile-displayname {
+        margin-top: 1rem;
+    }
+    div.cp-app-profile-link.hidden{
```
yflory, line 92: As mentioned in previous PRs, `hidden` is not a good class name as it could be used by libraries we embed in CryptPad and the style could leak to or from this library.

www/profile/inner.js
```diff
@@ -158,8 +175,11 @@ define([
         }, Messages.profile_addLink);
         APP.$linkEdit = $(button);
         $block.append(button);
-        var save = h('button.btn.btn-primary', Messages.settings_save);
-        var text = h('input');
+        var save = h('button.btn.btn-primary', Messages.settings_save, { 'aria-labelledby': 'cp-save-link' });
```
yflory, line 178: We usually put the attribute object before the content. This order should work but it is a bit inconsistent with the rest of the code.

www/profile/inner.js
```diff
@@ -158,8 +175,11 @@ define([
...
+        var text = h('input', { 
+            id: 'cp-save-link',
```
yflory, line 180: Same as before, `h('input#cp-save-link', { ... })`

www/profile/inner.js
```diff
@@ -221,14 +242,17 @@ define([
         if (friends[data.curvePublic]) {
             // Add friend message
             APP.$friend.append(h('p.cp-app-profile-friend', [
-                h('i.fa.fa-address-book'),
+                h('i', { class: 'fa fa-address-book', 'aria-hidden': 'true' }),
```
yflory, line 245: Same as before about ids and classes with hyperscript. This is especially true when you modify existing code since we usually want to keep the diff as small as possible!

www/profile/inner.js
```diff
@@ -261,10 +285,13 @@ define([
...
         var addCancel = function () {
-            var cancelButton = h('button.btn.btn-danger.cp-app-profile-friend-request', [
-                h('i.fa.fa-user-times'),
-                Messages.cancel
-            ]);
+            var cancelButton = h('button', 
+                { class: 'btn btn-danger cp-app-profile-friend-request', 'aria-labelledby': 'cp-cancel-button' }, 
+                [
+                    h('i', { class: 'fa fa-user-times', 'aria-hidden': 'true' }),
+                    h('span', { id: 'cp-cancel-button' }, Messages.contacts_remove || 'Cancel Request')
```
yflory, line 292: Same about hyperscript and don't hesitate to break the attribute object into multiple lines when it contains multiples elements.

And I'm not sure about the translation key here. If I understand/remember the code correctly, this is a button to cancel a contact request that has not been accepted yet. It was changed from "cancel" to "contacts_remove" but I don't understand why.
You should also remove the hardcoded value here: it's adding unnecessary bytes to the file and if you do so when adding a new key, we may skip that translation key and never notice that it's missing until we test with a different language. There may be some places in the code that still have such hardcoded translation but we want to avoid/remove them with our new processes.

www/profile/inner.js
```diff
@@ -326,10 +353,13 @@ define([
...
             if (isMuted) {
-                var unmuteButton = h('button.btn.btn-secondary.cp-app-profile-friend-request', [
-                    h('i.fa.fa-bell'),
-                    Messages.contacts_unmute || 'unmute'
-                ]);
+                var unmuteButton = h('button', 
+                    { class: 'btn btn-secondary cp-app-profile-friend-request', 'aria-labelledby': 'cp-unmute-button' }, 
+                    [
+                        h('i', { class: 'fa fa-bell', 'aria-hidden': 'true' }),
+                        h('span', { id: 'cp-unmute-button' }, Messages.contacts_unmute || 'Unmute')
```
yflory, line 360: In multiple places you added similar ids. I don't think they will create issue here but be careful with generic names, even when you use the `cp-` prefix.
"cp-cancel-button", "cp-unmute-button", etc. are names that could be used in the rest of the product and we could accidentally apply styles here when modifying something else. Usually we try to add the application in the name when describing a very specific element: "cp-profile-cancel-button"

www/profile/inner.js
```diff
@@ -374,7 +407,7 @@ define([
 
         var $delButton = $('<button>', {
             'class': 'cp-app-profile-avatar-delete btn btn-danger fa fa-times',
-            title: Messages.fc_delete
+            title: Messages.profile_remove_avatar //XXX
```
yflory, line 410: No need for XXX here if you've already added the temporary key to customize.dist/messages.js

www/profile/inner.js
```diff
@@ -463,6 +498,14 @@ define([
         editor.setOption('readOnly', false);
         cm.configureTheme(common, function () {});
 
+        editor.on("keydown", function (cm, event) {
```
yflory, line 501: CodeMirror directly includes a way to handle keyboard events:
```javascript
editor.setOption("extraKeys", {
    "Esc": function () {
        cm.getInputField().blur();
        $(save).focus();
    }
}
```
This would avoid any conflict with other shortcuts that can be added.

www/profile/inner.js
```diff
@@ -503,6 +547,11 @@ define([
         if (!APP.editor) { return; }
         APP.editor.setValue(data.description || "");
         APP.editor.save();
+        if ($.trim(descriptionData) === "") {
```
yflory, line 550: It's probably faster to use the native "trim" instead of jQuery: `descriptionData.trim()`

www/profile/inner.js
```diff
@@ -512,9 +561,12 @@ define([
         var $div = $(h('div.cp-sidebarlayout-element')).appendTo($container);
         APP.$edPublic = $('<button>', {
             'class': 'btn',
-        }).append(h('i.fa.fa-key'))
-          .append(h('span', Messages.profile_copyKey))
-          .click(function () {
+            'aria-labelledby': 'cp-copy-key-button'
+        }).append(
+            h('i', { 'class': 'fa fa-key', 'aria-hidden': 'true' })
+        ).append(
```
yflory, line 567: You can append multiple elements with jQuery using an array:
```javascript
$something.append([
  element1,
  element2
])

File: www/profile/inner.js [hunk 21]
```diff
@@ -535,15 +587,15 @@ define([
...
-        APP.$copyData = $(h('button.btn.btn-secondary', [
-            h('i.fa.fa-clipboard'),
-            h('span', Messages.support_copyUserData)
-        ])).click(function () {
+        var $block = $('<div>', {class:LINK_ID}).appendTo($container);
+        APP.$copyData = $(h('button', 
+            { class: 'btn btn-secondary', 'aria-labelledby': 'cp-copy-data-button' }, 
+            [   h('i', { class: 'fa fa-clipboard', 'aria-hidden': 'true' }), 
+                h('span', { id: 'cp-copy-data-button' }, Messages.support_copyUserData)
+            ])).click(function () {
```
yflory, line 595: Multiple indentations and coding style inconsistencies here
```
APP.$copyData = $(h('button.btn.btn-secondary', {
    'aria-labelledby': '...'
}, [
  h('i.fa.fa-clipboard', { ... }),
  h('span#cp-copy-data-button', ...)
]).click(function () {
```

yflory (changes requested):
Most of the comments in this review are about coding styles consistency. I didn't see anything wrong with the code itself and it should work as expected.

While testing the PR, I found one issue and I have one question.
**Issue**: on small screen, the "link" input is too large and the save button is outside the screen.
**Question/comment**: Why did you split the "view" and "share" buttons and also the rendered description and "edit" description?

yflory (approved):

Issue #1722, opened 2024-11-26
 Long question titles overflow inside Forms


dariiing:
When the title of a question is longer than expected, it extends beyond the designated area, disrupting the layout and overall user experience.
![Screenshot 2024-11-26 123242](https://github.com/user-attachments/assets/ce460c09-c8b2-4f18-bafb-33069a742cfd)

davidbenque:
I think this is partly a wrapping issue as your example shows repeated characters with no space. But there are other issues too such as the "required" label height.

PR #1747, opened 2024-12-18
Fix long question text overflow and weird sizing of the required label in forms

DianaXWiki:
This PR aims to fix the overflowing of long questions in Forms app and the weird sizing of the Required button of a question, mentioned in [#1722 ](https://github.com/cryptpad/cryptpad/issues/1722)

### DIFF
--- a/www/form/app-form.less
+++ b/www/form/app-form.less
@@ -613,10 +613,9 @@
                             margin-top: -10px;
                             margin-right: -10px;
                             &.cp-is-empty {
-                                max-height: fit-content;
+                                height: fit-content;
                                 padding: 3px;
-                                border: 2px solid @cp_alerts-danger-fg;
-                                border-radius: @variables_radius;
+                                border-radius: 0 @variables_radius 0 0;
                                 background-color: @cp_alerts-danger-bg;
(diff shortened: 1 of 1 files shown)


dariiing (approved):

davidbenque:
There is still an issue with the required label growing to match the height of the question

image

www/form/app-form.less
```diff
@@ -605,15 +605,20 @@
...
                         .cp-form-required-tag {
                             padding: 5px;
                             margin-top: -10px;
                             margin-right: -10px;
                             &.cp-is-empty {
+                                max-height: fit-content;
                                 padding: 3px;
-                                border: 2px solid @cryptpad_color_red;
-                                color: @cp_form-invalid;
+                                border: 2px solid @cp_alerts-danger-fg;
```
davidbenque, line 618: Border can be removed

www/form/app-form.less
```diff
@@ -605,15 +605,20 @@
...
+                                border-radius: @variables_radius;
```
davidbenque, line 619: Border radius should be none except at the top-right where the question container is rounded. So `0 @variables_radius 0 0`

PR #1751, opened 2024-12-26
Use shallow fetch to upgrade instance

nisbet-hubbard:
In the vast majority of cases, admins don’t need every version of every file since last upgrade. So a shallow fetch would be more ideal.

This PR accompanies https://github.com/cryptpad/documentation/pull/103.


davidbenque:
Hi @nisbet-hubbard sorry our `staging` branch wasn't up to date, could you rebase again?

Issue #485, opened 2020-01-07
Setting to disable opening of new windows/tabs


7adietri:
I would like to see a setting that lets me disable the CryptPad behaviour of opening a new window/tab when I open a pad or the teams list. Or put in another way, let me stay in one tab, no matter what I do.

Everything else in CryptPad is pretty awesome, thanks!

fsantiago07044:
I second this. That’s quite annoying, but I can stomach it if there’s a valid technical or security based reason.

ansuz:
If you're browsing with [a browser that support SharedWorkers](https://caniuse.com/#feat=sharedworkers) then you should notice that your second tab opens much more quickly than your first. This is because your account is already connected to the server in the background thanks to your first tab having launched a worker process to which all new tabs can connect.

This also means that all your CryptPad traffic for that browser runs over one websocket connection instead of as sockets as there are tabs. This incurs significantly less load on our server.

Shared workers shut down as soon as the last tab that required them closes. Opening a new tab is an easy way to ensure that the worker doesn't shut down and restart as you leave your first page and open your second in the same tab.

7adietri:
Thanks for the explanation! I'm often navigating from `CryptDrive` to `Teams` (and back), which also opens a new tab every time. Is this due to worker performance as well, or could it be done in the same tab?

AInteriorB:
Just found this thread via Google. I can confirm, after navigating through three documents I have six open tabs. Please provide a config option, where we can disable this behavior. That would be great.

roaldarbol:
This would be great. I use CryptPad in Brave, and as a Chromium browser it is possible to create a "shortcut", a dockable web app (e.g. that's how I use the ProtonMail web-interface as a desktop application). Currently, this is possible with CryptPad (CryptDrive) too, but every time you open a new document it opens a new pane in a(nother) web browser. 

If it was possible to disable the pop-up behaviour, then CryptPad could be a fully functional desktop app with no added effort. 
<img width="218" alt="Screenshot 2022-05-15 at 17 47 31" src="https://user-images.githubusercontent.com/25629697/168484214-d4ccbd61-c93f-43fa-a259-362cfa70d89a.png">

zen2:
+10

We are testing CryptPad among others solutions and it fit globally our need but the "new tab" behaviour make it not usable practically. You  quickly finish with dozens of open tab and all testers have describe this behaviour as too much annoying for production usage.

roaldarbol:
[...]

PR #1752, opened 2025-01-06
Treat relative URLs in the sandbox as relative to the outer domain

ansuz:
I believe that there used to be some event handlers for `<a>` tags in the sandbox which ensured that relative URLs were opened under the main/outer domain rather than the sandbox/inner one. This seems to have stopped working in one of the recent releases. I only noticed when I opened an old markdown document which used relative links between different pads.

I realized that this functionalty could be re-implemented much more simply by using a [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) tag. Adding such a tag with an `href` attribute pointing to the outer domain causes all relative links to be interpreted as relative to that URL, rather than to the sandbox (which is unlikely to be anyone's intended behaviour).

Types of URLs other than relative ones should not be affected, so I think this is unlikely to cause any new problems. That said, I haven't tested it very thoroughly except to confirm that relative links open in the expected context.

On a side note, a base tag is also a convenient way to define a global `target` behaviour, which could be relevant for #485.

### DIFF
--- a/www/common/sframe-common.js
+++ b/www/common/sframe-common.js
@@ -1057,6 +1057,11 @@ define([
 
             Mailbox.create(funcs);
 
+            // automatically configure all relative links in the inner iframe
+            // to point to the outer domain by adding a 'base' element to iframe's <head>
+            // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base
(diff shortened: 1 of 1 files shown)


davidbenque:
Thank you @ansuz !

ansuz:
:tada:

PR #1753, opened 2025-01-06
Buttons for moving question blocks in Forms (mobile)

### LINKED ISSUE #1587: Reorder questions should be doable without having to drag and drop
Moving questions (for example duplicated ones) within a form should be possible without requiring to "drag and drop" the question.

Ideally, there would be two arrows somewhere on the interface, which would allow to change the position of the question, "up" or "down".

zuzanna-maria:
Fixes #1587 (Form question blocks now have up/down buttons on mobile, allowing re-ordering without drag and drop)

### DIFF
--- a/www/form/app-form.less
+++ b/www/form/app-form.less
@@ -391,6 +391,26 @@
                 display: flex;
                 flex-flow: column;
                 flex: 1 1 1000px;
+                &.cp-no-drag {
+                    .cp-form-block-arrows {
+                        justify-content: space-between;
+                        max-width: fit-content;
+                        margin-inline: auto;
+                        padding: 6px;
+                        margin-top: -35px;
+                        border-radius: 0px @variables_radius 0px 0px;
+                        display: flex !important;
+                        column-gap: 20px;
+                        .cp-form-arrow {
+                            background-color: transparent;
+                            border: none;
+                            color: @cryptpad_text_col;
+                        }
+                    }
+                    .fa-ellipsis-h {
+                        display: none !important;
+                    }
+                }
 
                 .cp-form-creator-add-inline {
                     display: flex;
@@ -610,19 +630,7 @@
                         }
                     }
                     .cp-form-block-arrows {
-                        justify-content: space-between;
-                        max-width: fit-content;
-                        margin-inline: auto;
-                        padding: 6px;
-                        margin-top: -35px;
-                        border-radius: 0px @variables_radius 0px 0px;
-                        display: flex;
-                        column-gap: 20px;
-                        .cp-form-arrow {
-                            background-color: transparent;
-                            border: none;
-                            color: @cryptpad_text_col;
-                        }
+                        display: none;
                     }
 
                     &.sortable-ghost { visibility: hidden; }
--- a/www/form/app-form.less
+++ b/www/form/app-form.less
@@ -972,13 +972,14 @@
             background: fade(@cryptpad_text_col, 10%);
         }
         .cp-form-block-arrows {
-
-            float: left;
+            max-width: fit-content;
+            margin-inline: auto;
             padding: 5px;
             margin-top: -35px;
-            margin-left: -10px;
             border-radius: 0px @variables_radius 0px 0px;
-            .direction-arrow { margin: 5px; }
+            .direction-arrow-down {
+                margin-left: 20px; 
+            }
             background: fade(@cryptpad_text_col, 10%);
         }
     }
--- a/www/form/app-form.less
+++ b/www/form/app-form.less
@@ -207,16 +207,22 @@
                 width: 30px;
                 height: 30px;
                 border: none;
-                border-radius: 5px 0px 0px 5px;
+                border-radius: @variables_radius 0px 0px @variables_radius;
                 background-color: @cp_form-bg1;
                 color: @cryptpad_text_col; 
-                &.toggle-active {
-                    color: @cp_form-bg1 !important;
+                @media (max-width:723px) and (min-width:693px), (max-width:553px) and (min-width:524px) {
+                    border-radius: @variables_radius @variables_radius 0px 0px;
+                }
+                &.cp-toggle-active {
+                    color: @cp_form-bg1;
                     background-color: @cryptpad_text_col;
                 }
             }
-            #toggle-drag-on {
-                border-radius: 0px 5px 5px 0px !important;
+            #cp-toggle-drag-on {
+                border-radius: 0px @variables_radius @variables_radius 0px !important;
+                @media (max-width:723px) and (min-width:693px), (max-width:553px) and (min-width:524px) {
+                    border-radius: 0px 0px @variables_radius @variables_radius !important;
+                }
             }
 
             @media screen and (max-width: 1000px) {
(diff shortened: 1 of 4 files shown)


dariiing (changes requested):
Looking at the initial issue, reordering questions without drag and drop should be available on all devices, which would be particularly helpful for keyboard users.  This should be expanded to work across all devices.
(Also I'm not sure how to open the arrow functionality, but this might be an error on my side tbc)

www/form/inner.js
```diff
@@ -4020,11 +4020,9 @@ define([
             var changeType;
             if (editable) {
                 // Drag handle
-                dragHandle = h('span.cp-form-block-drag-handle', [
-                    h('i.fa.fa-ellipsis-h'),
-                    h('i.fa.fa-ellipsis-h'),
-                ]);
-
+                let drag = $('#toggle-drag-off').attr('class').indexOf('toggle-active') !== -1 ? false : true;
+                var dragEllipses = drag ? [h('i.fa.fa-ellipsis-h'), h('i.fa.fa-ellipsis-h')] : undefined;
```
yflory, line 4024: Removing the drag handle from DOM when drag is disabled means you'll need to redraw the entire page with all the questions whenever the user switches the mode. On mobile, it could take an unnecessarily "long" time for this.
A better way to improve performances on mobile could be to add both the drag handle and the arrows to the DOM, but add `display: none` to the one that shouldn't be visible.
You'd only have to update a class of a top-level container to switch which action is visible on all the questions at once, without redrawing the page.

www/form/inner.js
```diff
@@ -4260,12 +4258,49 @@ define([
                     }
                 }
             }
+            let shiftButtons;
+            let drag;
+            if (editable) {
```
yflory, line 4263: Here you're closing an `if (editable) {` to re-open one just after. Why not add your code to the existing `if` block?

www/form/inner.js
```diff
@@ -4260,12 +4258,49 @@ define([
...
+                drag = $('#toggle-drag-off').attr('class').indexOf('toggle-active') !== -1 ? false : true;
```
yflory, line 4264: If you merge your `if` block with the existing one, the `drag` state is already defined above

www/form/inner.js
```diff
@@ -4260,12 +4258,49 @@ define([
...
+            if (!drag) {
```
yflory, line 4265: Indentation error for this entire `if` block

www/form/inner.js
```diff
@@ -4260,12 +4258,49 @@ define([
...
+                    'aria-hidden': true
+                }));
+                var shiftBlock = function(direction) {
+                    var blockIndex = content.order.indexOf(uid);
+                    if (direction === 'up' && blockIndex > 0) {
+                        content.order.splice(blockIndex-1, 0, content.order.splice(blockIndex, 1)[0]);
+                    } else if (direction === 'down' && blockIndex < content.order.length-1) {
+                        content.order.splice(blockIndex+1, 0, content.order.splice(blockIndex, 1)[0]);
+                    }
+                    updateForm(framework, content, true);
```
yflory, line 4281: Since the "up" button is still visible on the first element and "down" button visible on the last one, it means we can click without triggering a change. In this case, it would be nice to avoid calling "updateForm" which, as mentioned earlier, can take some time on mobile with a big form.

www/form/inner.js
```diff
@@ -5099,9 +5136,30 @@ define([
                 editableStr
             ]);
 
+            var toggleOffclass = 'ontouchstart' in window ? 'toggle-active' : 'toggle-inactive'; 
+            var toggleOnclass = 'ontouchstart' in window ? 'toggle-inactive' : 'toggle-active'; 
+            var toggleDragOff = h(`button#toggle-drag-off.cp-form-view-drag.${toggleOffclass}.fa.fa-arrows`, {'aria-hidden': true, 'title': Messages.toggleArrows});
+            var toggleDragOn = h(`button#toggle-drag-on.cp-form-view-drag.${toggleOnclass}.fa.fa-hand-o-up`, {'aria-hidden': true, 'title': Messages.toggleDrag});
+            $(toggleDragOff).click(function() {
+                $(toggleDragOff).attr('class').indexOf('toggle-inactive') !== -1 ? $(toggleDragOff).toggleClass('toggle-inactive').toggleClass('toggle-active') && $(toggleDragOn).toggleClass('toggle-active').toggleClass('toggle-inactive') : undefined;
```
yflory, line 5144: You should try as much as possible to avoid long lines. Ideally, they should be shorter than 80 characters. In this case, using a ternary with an "undefined" at the end makes it all very difficult to read instead of a simple `if` statement.
I also don't think you need two classes here. If a button doesn't have the "active" class, it means it's "inactive"!

www/form/inner.js
```diff
@@ -5099,9 +5136,30 @@ define([
...
+                updateForm(framework, content, true);
+                APP.mainSortable.options.disabled =  true;
```
yflory, line 5146: Here, if you keep redrawing everytime with "updateForm", you probably don't need to update `APP.mainSortable` (it would be recreated anyway). If you update the code to avoid redrawing the entire form everytime, you need to replace the `updateForm` line above.

www/form/inner.js
```diff
@@ -5099,9 +5136,30 @@ define([
...
+            });
+            $(toggleDragOn).click(function() {
+                $(toggleDragOn).attr('class').indexOf('toggle-inactive') !== -1 ? $(toggleDragOn).toggleClass('toggle-inactive').toggleClass('toggle-active') && $(toggleDragOff).toggleClass('toggle-active').toggleClass('toggle-inactive') : undefined;
+                updateForm(framework, content, true);
+                APP.mainSortable.options.disabled =  false;
```
yflory, line 5151: Since this is almost exactly the same block as above, you can probably make it a function:
```javascript
const updateDrag = state => {
  return function () {
    $dragOn.toggleClass('active', state);
    $dragOff.toggleClass('active', !state);
    APP.mainSortable.options.disabled = !state;
    // and update the container class to show either the drag handle or the arrows
  };
};
$dragOn.click(updateDrag(true));
$dragOff.click(updateDrag(false));

File: www/form/inner.js [hunk 9]
```diff
@@ -5099,9 +5136,30 @@ define([
...
+
+            var drag = h('div', [
+                toggleDragOff,
+                toggleDragOn
+            ]);
+
             return [
                 preview,
                 previewSettings,
+                drag,
```
yflory, line 5162: A small issue about the style on mobile, this could also probably be fixed in the .less files directly:
image

yflory (changes requested):

yflory (approved):

www/form/app-form.less
```diff
@@ -203,6 +203,28 @@
...
             overflow: auto;
 
+            .cp-form-view-drag {
+                width: 30px;
+                height: 30px;
+                border: none;
+                border-radius: @variables_radius 0px 0px @variables_radius;
+                background-color: @cp_form-bg1;
+                color: @cryptpad_text_col; 
+                @media (max-width:723px) and (min-width:693px), (max-width:553px) and (min-width:524px), (max-width:379px) and (min-width:350px) {
```
yflory, line 213: I was waiting for @davidbenque to review the .less files, but this type of media-query seems very unreliable. If we make a small change to the rest of the UI, this may have to be changed. Additionally, it doesn't cover all cases if you enable the chat and/or the user list:
image

yflory (changes requested):

yflory (approved):

dariiing:
Testing:
- Weird double focus around new buttons form  (move up/down buttons)
- there are existing issues with the Form when using a screen reader, but newly added buttons are not announced correctly, or they are announced as "Button"

Issue #1741, opened 2024-12-16
Calendar title overflows inside calendar dropdown


dariiing:
### Description

Inside the Create/Edit Event modal, in the calendar dropdown, calendars with long titles overflow beyond the dropdown boundaries.
![Screenshot 2024-12-16 131524](https://github.com/user-attachments/assets/7cdc9b1a-8c14-490e-9add-4e46faf41acf)

### Steps to reproduce

1. Create a calendar with a very long title
2. Open the create/edit modal and change the calendar

### Expected behavior

Calendars with long titles should be displayed within the boundaries of the dropdown menu without causing layout issues.

### Which operating system are you using?

Microsoft Windows

### What browsers are you seeing the problem on?

Mozilla Firefox

### Extensions

.

### Version

2024.9.1

Issue #1553, opened 2024-07-10
Calendar dropdown not accessible 


dariiing:
![image](https://github.com/cryptpad/cryptpad/assets/95225431/3d04adc4-6000-4eec-a512-7dcfd9ab7249)
The calendar picker dropdown inside the event modal cannot be accessed using the keyboard. This limitation prevents users who do not use a mouse from selecting different calendars.

PR #1755, opened 2025-01-14
Improved calendar dropdown: keyboard accessibility and title overflow fixes

dariiing:
This PR is meant to fix calendar dropdown  fixes:
- #1553 , making the dropdown accessible by keyboard
- #1741,  prevents calendar titles from overflowing dropdown boundaries


yflory (changes requested):
I noticed a few issues while testing:
* If you open the dropdown and then click outside (anywhere else in the popup) to close it, the aria-expanded attributes stays "true"
* If I press "Space" on the button to open the dropdown, it doesn't focus the first element of the list as it does when I press "Enter"
* In our other dropdowns, I think we can press "Escape" to close the dropdown and focus its button. In this case, "Esc" completely closes the popup

yflory (approved):

PR #1759, opened 2025-01-22
Enable preview mode on mobile code app

DianaXWiki:


dariiing (commented):
From the accessibility standpoint, the changes are okay. However, there are existing issues that will need to be addressed:
- icons from buttons are missing the aria-hidden
- buttons are not correctly announced when using a screen reader
I would recommend them to be fixed in a separate PR because they cover more than the scope of this PR. I went ahead and created #1792 for future reference.

yflory (changes requested):
The code looks good, it seems you forgot to run the linter but it would be easy to fix, but I found some issues while testing.
* I always start in preview mode. It seems intentional since it's the first thing happening when creating a new document but if I open an "edit" link for a document, I expect the edit mode to be used first.
* If I am in "preview" mode and then I decide to switch to a new language (from Markdown to Python for example), I end up with no preview (because we don't have it for Python) and no editor also: it's just empty.
* If I have a Python document and decide to switch to Markdown, I end up with a split screen editor/preview, both sides being very narrow since I'm on mobile.

I haven't tested more than that because I feel like the way this PR is implemented is a bit too complex for what we're trying to fix. I think there should be an easy way to fix the issue with almost only CSS/LESS changes.
We already have a CSS class `.cp-app-code-fullpage` added to the container when the preview is disabled. With CSS media queries, we could have special style for small screens where we show either the editor or the preview depending on whether this class is applied or not. It would probably still requires some changes to the JavaScript to fix the initial state and issues with other languages though.

yflory (approved):

PR #1774, opened 2025-02-05
Add Skip Link

### LINKED ISSUE #1202: Link to skip the toolbar
> A skiplink should be provided to skip content that is repeated on multiple pages. In this case, it is a bit harder to decide whether the bar containing the logo and three buttons is ‘repeated content’ because of the title that differs on the pages. However, it would still be good for the usability and accessibility to provide a link that skips this bar. This link can be visually hidden and shown only when it receives the focus. It should be the first link on the page.

dariiing:
This PR introduces a skip link to improve accessibility, allowing users to quickly navigate to the main content. 
Fixes #1202

### DIFF
--- a/customize.dist/messages.js
+++ b/customize.dist/messages.js
@@ -118,6 +118,7 @@ define(req, function(AppConfig, Default, Language) {
 
     Messages._languages = map;
     Messages._languageUsed = language;
+    Messages.skipLink = "Skip to main content"; // XXX
 
     // Get keys with parameters
     Messages._getKey = function (key, argArray) {
--- a/customize.dist/src/less2/include/toolbar.less
+++ b/customize.dist/src/less2/include/toolbar.less
@@ -448,10 +448,11 @@
             .cp-toolbar-skip-link {
                 position: absolute;
                 top: -100px;
-                left: 0;
+                left: 45%;
                 background-color: @cryptpad_color_brand;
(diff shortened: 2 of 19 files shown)


davidbenque:
In Kanban my focus is on the first column's edit button by default instead of before the skip link

davidbenque:
In Rich Text pressing enter when the skip link is in focus doesn't produce any result

yflory (approved):

zuzanna-maria:
A description of some behaviours I found which may or may not be desired (Firefox desktop):
- In Kanban, skip link skips tag button + tag/drag toggles as well as generic main toolbar
- In Pad + Presentation, can’t escape editor with ESC in order to tab through interface - have to click outside of the editor manually; editor toolbars skipped		
- In diagram - focus gets ‘lost’ after navigating through right hand side panel; skip link doesn’t appear unless interface manually clicked first outside of the editor
- In document - OO embedded file menu/toolbar inaccessible to keyboard
- For multiple apps, the ‘do you want to store’ modal is the first thing that is accessed through tabbing, not the skip link

davidbenque:
Many thanks @zuzanna-maria for catching these. I think the majority are known issues which could not be resolved for this v1 of the skip link. Pinging @dariiing anyway as the last one about storing dialog may not be known.

PR #1775, opened 2025-02-07
Prevent saving file to Drive multiple times

### LINKED ISSUE #1745: Files can be stored in CryptDrive multiple times
### Description

The 'Store in CryptDrive' function creates multiple Drive shortcuts for the same file

### Steps to reproduce

1. Go to e.g. [cryptpad.fr/pad](http://cryptpad.fr/pad) and create a new Pad document
2. Click 'Store' on the popup that asks if file should be saved -- file appears in Drive
3. File menu > 'Store in CryptDrive' -- every time 'Store in CryptDrive' is clicked creates a new shortcut

### Relevant logs

_No response_

### Expected behavior

Only one shortcut per file should be created in the Drive -- clicking 'Store in CryptDrive' multiple times should produce an alert that file is already stored

_No response_

_No response_

### Extensions

n/a

### Version

2024.12.0

zuzanna-maria:
- Fixes #1745 (can no longer save file to drive if already saved)

### DIFF
--- a/www/common/common-ui-elements.js
+++ b/www/common/common-ui-elements.js
@@ -594,16 +594,19 @@ define([
         let $icon = $button.find('> i');
 
         let attributes = {};
+        let style = {};
         let btnClass = $button.attr('class');
         let btnId = $button.attr('id');
         let btnTitle = $button.attr('title');
+        let btnVisibility = $button.attr('style');
         if (btnClass) { attributes['class'] = btnClass; }
         if (btnId) { attributes['id'] = btnId; }
         if (btnTitle && !attributes.title) { attributes['title'] = btnTitle; }
-
+        if (btnVisibility) { attributes['style'] = btnVisibility; }
         return UIElements.createDropdownEntry({
             tag: 'a',
             attributes: attributes,
+            style: style,
             content: [
                 h('i',{ 'class': $icon.attr('class') }),
                 h('span', $button.text())
@@ -865,7 +868,9 @@ define([
                 if (callback) { button.click(callback); }
                 break;
             case 'storeindrive':
-                button = $(h('button.cp-toolbar-storeindrive', [
+                button = $(h('button.cp-toolbar-storeindrive', {
+                    style: 'display:none;'
+                }, [
                     h('i.fa.fa-hdd-o'),
                     h('span.cp-toolbar-name.cp-toolbar-drawer-element', Messages.toolbar_storeInDrive)
                 ])).click(common.prepareFeedback(type)).click(function () {
@@ -3347,9 +3352,7 @@ define([
         if (data && data.stored) {
             if (!data.inMyDrive) {
                 $('.cp-toolbar-storeindrive').show();
-            } else {
-                $('.cp-toolbar-storeindrive').hide();
-            }
+            } 
             return;
         }
         var priv = common.getMetadataMgr().getPrivateData();
--- a/www/common/common-ui-elements.js
+++ b/www/common/common-ui-elements.js
@@ -865,9 +865,7 @@ define([
                 if (callback) { button.click(callback); }
                 break;
             case 'storeindrive':
-                button = $(h('button.cp-toolbar-storeindrive', {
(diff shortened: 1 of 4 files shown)


www/common/common-ui-elements.js
```diff
@@ -594,16 +594,19 @@ define([
...
+        let btnVisibility = $button.attr('style');
         if (btnClass) { attributes['class'] = btnClass; }
         if (btnId) { attributes['id'] = btnId; }
         if (btnTitle && !attributes.title) { attributes['title'] = btnTitle; }
-
+        if (btnVisibility) { attributes['style'] = btnVisibility; }
         return UIElements.createDropdownEntry({
             tag: 'a',
             attributes: attributes,
+            style: style,
```
yflory, line 609: I'm not sure I understand why we have a "style" element here if we already have `attributes['style']` above. It seems like one will override the other.

www/common/common-ui-elements.js
```diff
@@ -3349,7 +3352,7 @@ define([
         if (data && data.stored) {
             if (!data.inMyDrive) {
                 $('.cp-toolbar-storeindrive').show();
-            }
+            }
```
yflory, line 3355: "Space" character to remove

www/common/drive-ui.js
```diff
@@ -5094,9 +5094,9 @@ define([
                 sframeChan.query('Q_STORE_IN_TEAM', {
                     href: data.href || data.rohref,
                     password: data.password,
-                    path: paths[0].path[0] === 'template' ? ['template'] : undefined,
                     title: data.title || '',
-                    teamId: -1
+                    teamId: -1,
+                    path: paths[0].path[0] === 'template' ? ['template'] : undefined
```
yflory, line 5099: Try to keep the original order to avoid unnecessary diff :)

www/common/outer/async-store.js
```diff
@@ -1220,7 +1221,20 @@ define([
                 if (!s.id) { inMyDrive = res.length; }
 
                 Array.prototype.push.apply(allData, res);
+
+                if (res && res[0]) {
+                    var elementPath = store.manager.user.userObject.findFile(res[0].id)[0];
```
yflory, line 1226: Multiple issues here:
1. "res" may reference data from a team drive but you're trying to search the path from your own drive
2. "res" is an array because the pad may be stored in your drive multiple times. You're pushing to the array only the path of the first occurence found

www/common/outer/async-store.js
```diff
@@ -1220,7 +1221,20 @@ define([
...
+                    if (elementPath) {
+                        path.push(elementPath);
+                    }
+                }    
+
             });
+
+            if (path[0]) {
```
yflory, line 1234: I don't really understand why `path` should be an array in you only use the first entry

www/common/outer/async-store.js
```diff
@@ -1287,6 +1300,8 @@ define([
                     });
                     return;
                 }
+            } else if (inTrash) {
+                store.manager.user.userObject.restore(path[0]);
```
yflory, line 1304: If I understand correctly (I didn't test), it seems this is going to restore any pads from your drive that you open. This function is called everytime someone opens a document, updates its title or tries to store it. This isn't something we really want (trashed document should stay there unless there is an explicit action to restore them).

yflory (changes requested):
The latest changes seem a bit out of scope for this PR. They aren't about preventing multiple save anymore but about restoring documents from the trash.
The initial goal of the PR was to fix a clear issue, but the part about the trash may change the user experience and probably requires further discussions about how to handle this issue.

www/common/outer/async-store.js
```diff
@@ -1253,7 +1253,7 @@ define([
                 }
                 obj.userObject.setHref(channel, null, href);
             });
-
+
```
Chouhartem, line 1256: trailing spaces to remove

yflory (approved):

Issue #1764, opened 2025-01-31
Lack of error description [Settings Admin]


dariiing:
When attempting to upload a new custom logo, the generic "Error" message is displayed without any indication of what went wrong. In this case, I discovered that the uploaded file was too large only after examining the console.

Issue #1763, opened 2025-01-31
Lack of error description


dariiing:
When attempting to submit a form without picking an option on how to answer the form, you get the generic "Error" message.

dariiing:
> > When attempting to submit a form without picking an option, you get the generic "Error" message.
> 
> [@dariiing](https://github.com/dariiing) can you specify which form?

Hello,
This occurs with any type of form.

Issue #1762, opened 2025-01-31
Lack of error description [Automatic download limit]


dariiing:
Found by `HAN scan 08-24`:

In the ‘automatic download limit’ setting, if letters are entered instead of numbers, a status message containing the text ‘error’ appears on the screen. This message lacks any descriptive text, providing no information about the nature of the error or how to correct it. To resolve this, the error message should be more descriptive, clearly indicating the issue (e.g., ‘Please enter a number between 1-100’) to help users correct their input.

PR #1789, opened 2025-02-18
Add descriptive error messages

dariiing:
This PR adds descriptive error message in multiple areas:
-  #1762 
-  #1763
-  #1764
-  #1790
-  #1791

### DIFF
--- a/customize.dist/messages.js
+++ b/customize.dist/messages.js
@@ -118,7 +118,7 @@ define(req, function(AppConfig, Default, Language) {
 
     Messages._languages = map;
     Messages._languageUsed = language;
-    Messages.admin_logoSize_error = "The logo size must be smaller than 200KB"; // XXX
+    Messages.admin_logoSize_error = "The logo size is too large"; // XXX
     Messages.limit_error = "Please enter a valid number"; // XXX
     Messages.positiveNumber_error = "Please enter a positive number"; // XXX
     Messages.answerType_error = "Please select how to answer the form"; // XXX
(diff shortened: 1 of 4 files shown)


customize.dist/messages.js
```diff
@@ -118,6 +118,10 @@ define(req, function(AppConfig, Default, Language) {
 
     Messages._languages = map;
     Messages._languageUsed = language;
+    Messages.admin_logoSize_error = "The logo size must be smaller than 200KB"; // XXX
```
davidbenque, line 121: The value "200KB" should be passed from the code as a value, with the placeholder "{0}" in the translated string

www/admin/inner.js
```diff
@@ -884,6 +884,11 @@ define([
                     UI.warn(Messages.error);
                     return;
                 }
+                if (files[0].size > 200 * 1024) {
```
davidbenque, line 887: use a variable here, same one that should be passed to the message

customize.dist/messages.js
akbarjorayev, line 121:
```suggestion
    let maxLogoSize = 200;

    Messages._languages = map;
    Messages._languageUsed = language;
    Messages.admin_logoSize_error = `The logo size must be smaller than ${maxLogoSize}KB`; // XXX
```

www/admin/inner.js
akbarjorayev, line 887:
```suggestion
                let maxLogoSizeMb = 200 * 1024;
                if (files[0].size > maxLogoSizeMb) {
```

akbarjorayev (commented):
max file size variable added

yflory (approved):

Issue #1809, opened 2025-03-06
Tags are not accessible via keyboard


dariiing:
In Kanban, you cannot navigate through tags in order to filter the board using the keyboard.

PR #1846, opened 2025-03-28
Improve Kanban Tags accessibility with keyboard navigation

dariiing:
This PR addresses the Kanban tags and resolves the issue by making them accessible via keyboard, closing #1809.

### DIFF
--- a/www/kanban/inner.js
+++ b/www/kanban/inner.js
@@ -1153,11 +1153,6 @@ define([
                         'data-tag': t,
                         'tabindex': 0
                     }, t));
-                    $(tag).keydown(function (e) {
-                        if (e.which === 13 || e.which === 32) {
-                            $(tag).click();
-                        }
-                    });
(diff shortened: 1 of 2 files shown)


www/kanban/inner.js
```diff
@@ -1150,8 +1150,14 @@ define([
                 allTags.forEach(function (t) {
                     var tag;
                     $list.append(tag = h('span', {
-                        'data-tag': t
+                        'data-tag': t,
+                        'tabindex': 0
                     }, t));
+                    $(tag).keydown(function (e) {
```
yflory, line 1156: Small optimization possible here. You can use the `$tag` variable defined just after to avoid initializing a jQuery object twice:
```javascript
var $tag = $(tag).click(function () {
    // ...
}).keydown(function (e) {
    // Use $tag here instead of $(tag)
});

yflory (changes requested):

Issue #1781, opened 2025-02-11
Lack of clear focus indicators and focusable elements on the Notification page


dariiing:
On https://cryptpad.fr/notifications/, elements don't have a clear focus indicator and other buttons are not focusable:

For example: In the **History** section, I can navigate through notifications using the Tab key, but the focus indicator is unclear, it's just a blue line. Additionally, the "Load More History" button is not accessible via keyboard.

![Image](https://github.com/user-attachments/assets/c1bf7e2f-f060-4db5-8362-c5665275790d)

PR #1848, opened 2025-03-28
Improve notification page accessibility, responsiveness and UI

dariiing:
This PR improves the notification page by addressing multiple UI and accessibility issues, closing #1781:
- make all the buttons and clickable elements reachable via keyboard
- enhance responsiveness for smaller screens
- refine notification UI
- add screen reader support
- fix contrast ratio on elements

Tested with NVDA & Axe Core.


www/common/sframe-common-mailbox.js
```diff
@@ -81,8 +81,18 @@ define([
...
+                    title: Messages.user_profile,
+                    'aria-label': Messages.user_profile,
+                    role: 'button'
+                });
                 Common.displayAvatar($(avatar), userData.avatar, userData.displayName || userData.name);
+                $(avatar).keydown(function (e) {
+                    if (e.which === 13 || e.which === 32) {
+                        $(avatar).click();
+                    }
+                });
```
yflory, line 95: You can use `Util.onClickEnter($(avatar), handler, { space: true });` to avoid triggering a click on Enter or Space

customize.dist/src/less2/include/avatar.less
```diff
@@ -26,7 +26,6 @@
 & {
     .avatar_vars();
     &.cp-avatar {
-        overflow: hidden;
```
yflory, line 29: What was the issue with `overflow: hidden`? I can't really see a use-case for having the avatar bigger than its container

www/notifications/inner.js
```diff
@@ -154,6 +161,11 @@ define([
             }
         });
 
+        $(dismissAll).keydown(function (e) {
+            if (e.keyCode === 13 || e.keyCode === 32) {
+                $(dismissAll).click();
+            }
+        });
```
yflory, line 168: Same as before: `Util.onClickEnter`

yflory (changes requested):

yflory (approved):

PR #1913, opened 2025-05-23
Improve screen reader support for Document access

dariiing:
When a link to a document is received, users can either open the document or see a status/error message (eg: 'Document was destroyed by an owner'). These messages are now announced by screen readers. Also added visible outlines for links as a minor enhancement.

### DIFF
--- a/customize.dist/loading.js
+++ b/customize.dist/loading.js
@@ -15,7 +15,7 @@ define([
 
     elem.innerHTML = [
         '<div class="cp-loading-logo">',
-            '<img class="cp-loading-cryptofist" src="/api/logo?' + urlArgs + '" alt="' + Messages.label_logo + '">',
+            '<img class="cp-loading-cryptofist" src="/api/logo?' + urlArgs + '" alt="" aria-hidden="true">',
         '</div>',
         '<div class="cp-loading-container">',
             '<div class="cp-loading-spinner-container">',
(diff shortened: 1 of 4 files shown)


www/code/inner.js
```diff
@@ -179,6 +179,9 @@ define([
 
 
     var isSmallScreen = () => window.innerWidth <= 600;
+    const showEditor = Util.once(() => {
+    $('#cp-app-code-editor').css('display', '');
```
yflory, line 183: Indentation issue

www/pad/inner.js
```diff
@@ -757,6 +757,7 @@ define([
         */
 
         framework.onEditableChange(function(unlocked) {
+            $('#cp-app-pad-editor').show();
```
yflory, line 760: This will break the "cache" system. While we wait for the content to be editable (meaning it is fully synced with the server), we display a read-only "cached" version of the document. This won't trigger en "editable change" event and will result in a page showing only the toolbar.

www/whiteboard/inner.js
```diff
@@ -474,6 +474,8 @@ define([
 
 
         framework.onEditableChange(function () {
+            $('#cp-app-whiteboard-canvas-area').show();
```
yflory, line 477: We may have the same issue than for Rich Text documents ("pad"), but I didn't test

yflory (changes requested):

yflory (approved):

Issue #1921, opened 2025-05-27
2025.3.1 update succeeds, but breaks when attempting to run.


thraeryn:
### Description

Updated CryptPad from 2025.3.0 to 2025.3.1. Update seemed successful. Starting service drove CPU cycles through the roof while making NGINX return a "502 Bad Gateway" error.

### Steps to reproduce

1. Update Cryptpad from 2025.3.0 to 2025.3.1, as described in release notes.
2. Start Cryptpad service.

### Relevant logs

```shell
ay 27 10:10:46 thraeryn.red systemd[1]: Started cryptpad.service - CryptPad API server.
May 27 10:10:51 thraeryn.red node[1485489]: [2025-05-27T15:10:51.243Z] UNCAUGHT EXCEPTION IN HTTP WORKER
May 27 10:10:51 thraeryn.red node[1485489]: TypeError [ERR_INVALID_ARG_TYPE]: The "paths[1]" argument must be of type string. Received undefined
May 27 10:10:51 thraeryn.red node[1485489]:     at new NodeError (node:internal/errors:405:5)
May 27 10:10:51 thraeryn.red node[1485489]:     at validateString (node:internal/validators:162:11)
May 27 10:10:51 thraeryn.red node[1485489]:     at Object.resolve (node:path:1097:7)
May 27 10:10:51 thraeryn.red node[1485489]:     at /var/www/cryptpad/node_modules/connect-gzip-static/lib/gzip-static.js:31:24
May 27 10:10:51 thraeryn.red node[1485489]:     at Array.forEach (<anonymous>)
May 27 10:10:51 thraeryn.red node[1485489]:     at createCaches (/var/www/cryptpad/node_modules/connect-gzip-static/lib/gzip-static.js:24:11)
May 27 10:10:51 thraeryn.red node[1485489]:     at async createMethods (/var/www/cryptpad/node_modules/connect-gzip-static/lib/gzip-static.js:43:3) {
May 27 10:10:51 thraeryn.red node[1485489]:   code: 'ERR_INVALID_ARG_TYPE'
May 27 10:10:51 thraeryn.red node[1485489]: }
May 27 10:10:51 thraeryn.red node[1485489]: TERMINATING
```

### Expected behavior

I expected the update to work.

### Which operating system are you using?

Linux/BSD/UNIX

### What browsers are you seeing the problem on?

Mozilla Firefox

### Extensions

not applicable

### Version

Other

federicofantini:
[...]

PR #1922, opened 2025-05-28
Quick fix: Disable express compression

wginolas:
https://github.com/cryptpad/cryptpad/issues/1921


ghost:
If I understand properly, the original issue was caused by the outdated Node.js version of the person reporting it. Hence, we can forget this and keep Express compression enabled?

wginolas:
> If I understand properly, the original issue was caused by the outdated Node.js version of the person reporting it. Hence, we can forget this and keep Express compression enabled?

Yes

PR #1934, opened 2025-06-01
Fix RTL Layout Issues for Right-to-Left Languages

Goudarz:
This PR resolves multiple right-to-left (RTL) layout issues in the project, ensuring proper text alignment and directional attributes for languages like Persian, Arabic, and etc. The changes include:

- Implementing automatic `dir="rtl"` attributes where necessary.
- Adjusting bidirectional text formatting for compatibility.
- Fixing CSS-related RTL inconsistencies for a seamless user experience.

### DIFF
--- a/customize.dist/messages.js
+++ b/customize.dist/messages.js
@@ -13,6 +13,7 @@ var map = {
     'es': 'Español',
     'es_CU': 'Español cubano',
     'eu': 'Euskara',
+    'fa': 'فارسی',
     'fi': 'Suomi',
     'fr': 'Français',
     //'hi': 'हिन्दी',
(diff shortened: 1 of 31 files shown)


dariiing (changes requested):
Hello! Thanks for taking the time to fix this.
While looking over this PR,  I noticed a few UI inconsistencies, particularly with the border radius of buttons. Please make sure the border radius is applied consistently across the interface.  I’ve attached screenshots for reference showing where I spotted some issues:
image
image
image

zuzanna-maria:
Hello @Goudarz,

First of all, thank you so much for this PR and for the improvements implemented since the last review! We really appreciate all the work you've done to make CryptPad compatible with RTL languages.

However, while testing the code, I have noticed several further UI issues in various places which are introduced by changing the direction of the text. These include elements overlapping, being cut off or misaligned, missing padding, as well as some leftover button radius issues. Unfortunately, we don't currently have the capacity to fix these ourselves or to provide a thorough, point-by-point review of all the stylistic improvements this PR requires. 

We would recommend that, in order to address these issues, you compare the UI in this PR side-by-side with what the CryptPad UI looks like in production, making sure to systematically check all of CryptPad's functionality (all apps, drive, teams, user settings, calendar, modals, etc.). This will help you to identify and fix any discrepancies in the positioning and layout of elements like buttons, banners, dropdowns etc. 

Thank you again for your time!

PR #1995, opened 2025-07-14
add a checkup page test for /upload-blob

ansuz:
Hi!

Some time after updating my instance to [2025.6.0](https://github.com/cryptpad/cryptpad/releases/tag/2025.6.0) I noticed that uploads stopped working.

I wrote this test to help debug the issue, and eventually realized that my NGINX config was trying (and failing) to serve `/upload-blob` as a static asset rather than proxying to the (new-ish) endpoint on the back end.

Anyone using the simplified NGINX config is unlikely to experience this issue, but it might be useful for validating more advanced configurations.

### DIFF
--- a/www/checkup/main.js
+++ b/www/checkup/main.js
@@ -1649,6 +1649,27 @@ define([
         });
     });
 
+    // confirm that POST requests to the `/upload-blob` endpoint
+    // return something other than a 404, which would probably indicate
+    // a reverse proxy misconfiguration
+    assert(function (cb, msg) {
+        msg.appendChild(h('span', [
+            `The server returned a 404 error when attempting to reach the `,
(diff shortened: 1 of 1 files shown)


cremesk:
hi,

looks like i have the same problem with the `example-advanced.nginx.conf`.
Even if the checkup shows the problem, the question is how to fix it.
Many thanks for helping

meaz:
Having the same issue with `example-advanced.nginx.conf` also.

yflory:
Hi, thanks for the PR!
I just updated the advanced nginx config example with the missing value: https://github.com/cryptpad/cryptpad/commit/1287d9efd4d2a6a2d9253f36c9a9a3f65ded7792

PR #2031, opened 2025-08-18
Fix calendar-picker dropdown styling in event modal

DianaXWiki:
This PR aims to fix some styling issues identified within the calendar dropdown:

- [x] Dropdown arrow is misaligned and not an icon (color is off as well)
- [x] Uneven margins between top sections elements
- [x] Border radius does not match for dropdown button and dropdown menu (also no rounded margins)
- [x] Font size must be similar across elements of dropdown
- [x] Uneven padding in the top sections
- [x] Hide description in detail pop-up when the body contains only whitespace

### DIFF
--- a/www/calendar/inner.js
+++ b/www/calendar/inner.js
@@ -2224,6 +2224,15 @@ APP.recurrenceRule = {
                 let $dropdownButton = $el.find('.tui-full-calendar-dropdown-button');
                 let $dropdownMenu = $el.find('.tui-full-calendar-dropdown-menu');
 
+                let updateDropdownArrow = function(isOpen) {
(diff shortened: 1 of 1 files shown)


www/calendar/inner.js
```diff
@@ -412,9 +412,9 @@ define([
             return `<div class="event-location"> ${location_icon.outerHTML} ${str} </div>`;
         },
         popupDetailBody: function(schedule) {
-            var str = schedule.body;
+            var str = schedule.body ? schedule.body.trim() : ""; //check also whitespace
             delete APP.eventBody;
-
+            if (!str) { return "";}
```
Chouhartem, line 417: This should not be necessary with the ternary test on line `415`.

PR #2049, opened 2025-09-16
Fix text-tools button visibility for Kanban card title editing

DianaXWiki:
This PR fixes the unintended display of the markdown text-tools button when editing a Kanban card title. The button should only appear when editing the card body, ensuring markdown responsiveness on mobile devices.

image


Chouhartem (approved):
Looks good to me.

PR #2087, opened 2025-10-29
Fix: Removed avatars are no longer shown in team roster

### LINKED ISSUE #2043: Profile picture remain shown in team roster after deletion
### Description

When deleting an existing profile picture, it remains shown in your team roster.

### Steps to reproduce

1. Be part of a team
2. Add an avatar to your profile
3. Remove the profile picture previously set
4. Witness it's still being shown in the team members list

### Relevant logs

```shell

```

### Expected behavior

The deleted profile picture shouldn't show up in the team roster. 

Apple macOS

Google Chrome

### Extensions

Not relevant

### Version

2025.6.0

DianaXWiki:
This PR aims to fix #2043:

- [x] After deletion, avatar should not appear in the list of members from existing teams
- [x] Also fixed for lingering avatars in contacts and modal pop-ups
- [ ] There is the same issue for notifications, but it is out of scope for this PR


yflory (changes requested):
This fix doesn't work. When you remove your avatar in your settings, this fix is only removing the avatar in the team for yourself, but the other team members will still see it.

PR #2092, opened 2025-11-04
Add support for required poll questions in forms

### LINKED ISSUE #2078: Questions of type "poll" in Forms cannot be set as required
### Description

Inside CryptPad Forms, questions of type "poll" don't have the option to be made required (or optional).

Image

### Steps to reproduce

1. Create a CryptPad Form
2. Create a "Poll" question
3. Click "Edit" for the question
4. There's no "Answer: Optional, Required" option available

### Relevant logs

```shell
-
```

### Expected behavior

"Answer: Optional, Required" option should be available.

Linux/BSD/UNIX

Mozilla Firefox

### Extensions

-

### Version

2025.9.0

DianaXWiki:
This PR aims to fix #2078:

- [x] Add optional/required radio to poll questions
- [x] Ensure that required poll questions display appropriate UI elements and validation
- [x] Make required logic more permissive than other form questions: the user has to interact at least once with the response options

### DIFF
--- a/www/form/app-form.less
+++ b/www/form/app-form.less
@@ -757,7 +757,6 @@
                             }
                             &.cp-is-empty {
                                 height: fit-content;
-                                padding: 3px;
                                 border-radius: 0 @variables_radius 0 0;
                                 background-color: @cp_alerts-danger-bg;
                                 color: @cp_alerts-danger-text;


dariiing:
I'm not sure if making poll questions required is necessary. I noticed that in the code, a question is only marked as answered if one of the options is marked as “Yes.”
In some cases, a user might not include a “Yes” option and only wants to provide “No” or “Maybe” answers. In that situation, the question appears incomplete and remains marked as required, even though the user has already provided valid options.

DianaXWiki:
> I'm not sure if making poll questions required is necessary. I noticed that in the code, a question is only marked as answered if one of the options is marked as “Yes.” In some cases, a user might not include a “Yes” option and only wants to provide “No” or “Maybe” answers. In that situation, the question appears incomplete and remains marked as required, even though the user has already provided valid options.

Yes, these aspects were discussed before starting the implementation. The idea was to let users decide whether they want to enforce a more meaningful response (instead of the default all “No” or “Maybe”), ensuring consistent behavior with the “required” flag used for the other question types.

To reduce confusion, we could indeed add a UI hint when the user selects the “Required” option, clarifying how this mechanism works.

dariiing:
> > I'm not sure if making poll questions required is necessary. I noticed that in the code, a question is only marked as answered if one of the options is marked as “Yes.” In some cases, a user might not include a “Yes” option and only wants to provide “No” or “Maybe” answers. In that situation, the question appears incomplete and remains marked as required, even though the user has already provided valid options.
> 
> Yes, these aspects were discussed before starting the implementation. The idea was to let users decide whether they want to enforce a more meaningful response (instead of the default all “No” or “Maybe”), ensuring consistent behavior with the “required” flag used for the other question types.
> 
> To reduce confusion, we could indeed add a UI hint when the user selects the “Required” option, clarifying how this mechanism works.

Yep, a UI hint could work both on the responder and the creator side

yflory (changes requested):
The code looks good but I found a small margin issue with the "required" tag (`margin-right: 0.5rem;` on the icon)

image

PR #2097, opened 2025-11-07
install-onlyoffice.sh: fix --check for new install_version function

martinetd:
Does what it says on the tin: this will allow to check we didn't forget updating deps in nixpkgs again (script broke in 2025.6.0)

Thanks!

### DIFF
--- a/install-onlyoffice.sh
+++ b/install-onlyoffice.sh
@@ -271,17 +271,27 @@ install_old_version() {
 }
 
 install_version() {
-    ensure_command_available curl
-    ensure_command_available sha512sum
-    ensure_command_available unzip
-
     local DIR=$1
     local VERSION=$2
     local HASH=$3
     local FULL_DIR=$OO_DIR/$DIR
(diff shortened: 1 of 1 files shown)


martinetd:
By the way I noticed the new `oldest_needed_version` prop added in 2025.6.0 -- could you confirm why older versions are/could be needed?

If one had an install with v7, does one need v7 until all documents that were created with v7 have been opened at least once with (and converted to) v8?

ghost:
Hello @martinetd,

> If one had an install with v7, does one need v7 until all documents that were created with v7 have been opened at least once with (and converted to) v8?

It does indeed.

martinetd:
Thank you for confirming! nixos never had any version of cryptpad with onlyoffice < v7 so we'll be able to remove anything older than that at least (and make v7 configurable), that's a great new knob.

wginolas (approved):
Thank you for your contribution!

Checked the diff and tested it. 👍

PR #2102, opened 2025-11-13
Update drive tree UI

dariiing:
Hello,
This PR updates the drive tree UI and modularizes the underlying code.


www/common/drive-ui.js
```diff
@@ -4551,178 +4590,211 @@ define([
...
+                currentPath: currentPath,
+                rootPath: path,
+                events: {
+                    onFolderClicked: function (clickedPath, event) {
+                        // Check for shared folder restrictions
+                        var pathToCheck = clickedPath.slice(1); // Remove 'root'
+                        if (pathToCheck.length > 0) {
+                            var checkRoot = manager.find([ROOT]);
+                            var firstKey = pathToCheck[0];
+                            if (checkRoot && checkRoot[firstKey]) {
```
yflory: It seems that this aprt will only check for "deleted" and "restricted" shared folders if they are directly in the drive root. Shared folders can be located inside other classic folders and we should also apply the check to them.
If you want to get the shared folder id, you can use something like:
```js
var sfId = manager.isInSharedFolder(clickedPath);
```

www/common/drive-ui.js
```diff
@@ -4551,178 +4590,211 @@ define([
...
+                var sfId = null;
+                var f = null;
+                var droppable = true;
+                
+                if (elPath.length > 1 && elPath[0] === 'root') {
+                    var firstKey = elPath[1];
+                    if (checkRoot && checkRoot[firstKey]) {
+                        sfId = manager.isSharedFolder(checkRoot[firstKey]) && checkRoot[firstKey];
+                        if (sfId) {
+                            f = folders[sfId];
```
yflory: Same as a previous comment, this is not a way to check if an element is a shared folder or to get a shared folder id.
When you only have a path, you can use `var sfId = manager.isInSharedFolder(path);`

src/common/proxy-manager.js
```diff
@@ -562,6 +573,9 @@ const factory = (UserObject, Util, Hash,
                 });
             }
         }).nThen(function () {
+            if (moveError) {
```
yflory: This looks good but if you want to catch all the errors, you should also handle the 2 cases of `copyFromOtherDrive` above. You may encounter similar errors there.

PR #2104, opened 2025-11-14
Hide History button if history disabled

### LINKED ISSUE #1986: `AppConfig.enableHistory = false;` doesn’t remove History button
### Description

The comment says that 

```js
// Set enableHistory to false to remove the "History" button in all the apps.
```

But the button is still there after setting this to false.

### Steps to reproduce

1. Add `AppConfig.enableHistory = false;` to `/customize/application_config.js`
2. Restart instance
3. Open an app
4. Tap File
5. History is still there between Save as template and Tags

### Relevant logs

```shell

```

### Expected behavior

Setting enableHistory to false removes the "History" button in all the apps.

Apple iOS

Apple Safari

### Extensions

N/A

### Version

2025.6.0

zuzanna-maria:
- fixes #1986 (hides History button if history is disabled both in CryptPad and OnlyOffice apps)

### DIFF
--- a/www/common/common-ui-elements.js
+++ b/www/common/common-ui-elements.js
@@ -896,16 +896,15 @@ define([
                 button = makeButton('print', 'cp-toolbar-icon-print', Messages.printButtonTitle2, Messages.printText);
                 break;
             case 'history':
-                if (!AppConfig.enableHistory) {
-                    button = $('<span>');
-                    break;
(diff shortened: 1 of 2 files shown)


wginolas (approved):
Reviewed and tested 👍

yflory (approved):

Issue #1704, opened 2024-11-02
Performance improvements to example-advanced.nginx.conf


nisbet-hubbard:
### Description

There’s still room for [example-advanced.nginx.conf](https://github.com/cryptpad/cryptpad/blob/main/docs/example-advanced.nginx.conf) to be tuned.

### Steps to reproduce

See https://github.com/cryptpad/cryptpad/blob/main/docs/example-advanced.nginx.conf

### Expected behavior

1. Configuring load balancing allows nginx worker processes to use a shared memory zone, eg

```Nginx
    upstream node_backend {
        zone upstreams 64K;
        server 127.0.0.1:3000 max_fails=1 fail_timeout=2s;
        keepalive 2;
    }
```
as documented in the official blog: https://www.f5.com/company/blog/nginx/avoiding-top-10-nginx-configuration-mistakes#upstream-groups

2. Currently, security headers such as `Content-Security-Policy`, `X-XSS-Protection`,`X-Content-Type-Options` and `Permissions-Policy` are set even on static (ie non-HTML) assets, for which they are not designed. This has no security benefits and removing them reduces some payload. 

### Extensions

N/A

### Version

2024.9.0

PR #2106, opened 2025-11-17
Performance improvements (3/3) to example-advanced.nginx.conf

nisbet-hubbard:
This is the final part of the patch series for #1704. It configures one upstream group for each endpoint to allow worker processes (set to `auto`, ie the number of cores, in [most](https://exampleconfig.com/default/nginx/etc-nginx-nginx-conf?os=ubuntu2404) [nginx](https://src.fedoraproject.org/rpms/nginx/blob/rawhide/f/nginx.conf#_6) [distributions](https://exampleconfig.com/default/nginx/etc-nginx-nginx-conf?os=debian13)) to share a memory zone.

Since the [F5 example configuration](https://web.archive.org/web/20251002103745/https://www.f5.com/company/blog/nginx/avoiding-top-10-nginx-configuration-mistakes#upstream-groups) is for the use case of a single node-based backend application, I’ve simply adapted it for cryptpad.

Note that it’s [recommended](https://web.archive.org/web/20251002103745/https://www.f5.com/company/blog/nginx/avoiding-top-10-nginx-configuration-mistakes#no-keepalives) to set `keepalive` to twice the number of servers listed in the `upstream{}` block (which doesn’t limit the total number of connections to upstream servers that a worker process can open), so when testing on a large instance, there’s no need to increase this parameter.

The original article by F5, which is the company behind nginx, is here: https://www.f5.com/company/blog/nginx/avoiding-top-10-nginx-configuration-mistakes#upstream-groups.

### DIFF
--- a/docs/example-advanced.nginx.conf
+++ b/docs/example-advanced.nginx.conf
@@ -391,6 +391,6 @@ server {
         # Set all the rules you composed above.
         add_header Content-Security-Policy "default-src 'none'; child-src $childSrc; worker-src $workerSrc; media-src $mediaSrc; style-src $styleSrc; script-src $scriptSrc; connect-src $connectSrc; font-src $fontSrc; img-src $imgSrc; frame-src $frameSrc; frame-ancestors $frameAncestors" always;
 
-        try_files /customize/www/$uri /customize/www/$uri/index.html /src/$uri /www/$uri /www/$uri/index.html /src/$uri /customize/$uri;
+        try_files /customize/www/$uri /customize/www/$uri/index.html /src/$uri /www/$uri /www/$uri/index.html /customize/$uri;
     }
 }
--- a/docs/example-advanced.nginx.conf
+++ b/docs/example-advanced.nginx.conf
@@ -8,6 +8,21 @@
 #   installation (http server by the Nodejs process). If you are using CryptPad
 #   in production and require professional support please contact sales@cryptpad.fr
 
+proxy_http_version 1.1;
+proxy_set_header   "Connection" "";
+
+upstream node_http {
+    zone node_http 64k;
+    server localhost:3000 max_fails=1 fail_timeout=2s;
+    keepalive 2;
+}
+
+upstream node_ws {
+    zone node_ws 64k;
+    server localhost:3003 max_fails=1 fail_timeout=2s;
+    keepalive 2;
+}
+
 server {
     listen 80;
     listen [::]:80;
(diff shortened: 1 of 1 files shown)


ghost (approved):

ghost:
Hello @nisbet-hubbard,

Thanks a lot for this third PR!

We've been discussing some edge case where the application would be overloaded and a connection to the websocket would take more than 2s to succeed. What would happen?

Will a 500 error be generated? Will the connection will be automatically tried again?

We are especially worried because we are starting to introduce rate-limiting to CryptPad, leveraging Nginx functionalities for now. Wouldn't your proposed changes interfere with the limitations we would like to enforce?

For your information, our current snippet look like this:
```nginx
limit_req_zone $binary_remote_addr zone=wslimit:20m rate=30r/m;

server {

    # ....

    location ^~ /cryptpad_websocket {
        limit_req zone=wslimit burst=5 nodelay;
        limit_req_status 429;

    # ....
}
```

Looking forward to your feedback!
Thanks again.

nisbet-hubbard:
@mathilde-cryptpad, thanks for reviewing the patch!

Each request gets processed in the following order:

① limit_req —> ② proxy_next_upstream —> ③ server localhost:3003

If a request gets rate-limited at step (1), it’s never going to reach the load balancing logic. The parameters `error timeout http_500` and `max_fails=1 fail_timeout=2s` only affect the interaction between (2) and (3), as follows:

- `error timeout http_500` says if the upstream Cryptpad instance, the one sitting at 3003, returns an error, timeout or the code 500, [the request should be passed to the next upstream server](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream).
- `max_fails=1 fail_timeout=2s` is like saying max_fails=1/2s. It means if 1 error, timeout or code 500 is returned by an upstream server within a 2-second period, [consider this upstream server unavailable for another 2 seconds](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#server) (which isn’t relevant for the current request).
- Since there’s only one upstream server, the error, timeout or code 500 will simply get passed to the client.

So `fail_timeout=2s` doesn’t limit how long a Cryptpad instance can take to return a request (that’s controlled by `proxy_connect_timeout`, which defaults to 60s), no retry takes place, and a 500 gets returned only if that’s the response of the upstream Cryptpad instance.

You can also try load-testing this setup on the staging site using a script like the one here: https://www.getpagespeed.com/server-setup/nginx/nginx-rate-limiting#:~:text=load%20testing%20tool

davidbenque:
Hi Many thanks again for this contribution. Since we recently made changes in this area we want to do a bit more testing before merging. I'll remove it from the release milestone due this week but we are still actively looking at it.

PR #2117, opened 2025-11-27
Fix text overflow issues on Ctrl+E modal

### LINKED ISSUE #2109: UI overflow in the new document creation popup in some languages
### Description

When opening the new document popup in French (for instance), some labels overflows as depicted below:

Image

(Text is « Importer des fichiers » and « Importer des dossiers » for file and folder imports)

### Steps to reproduce

1. Switch CryptPad to French 🐓
2. Open your CryptDrive™
3. Open the document creation popup with the [Ctrl+e](https://docs.cryptpad.org/en/user_guide/keyboard_shortcuts.html#general-shortcuts) command
4. Notice the issue

### Relevant logs

```shell

```

### Expected behavior

The text should not overflow or be ellispised?

Linux/BSD/UNIX

Mozilla Firefox, Chromium

### Extensions

Not relevant

### Version

2025.9.0

dariiing:
Hello,
This fixes #2109

### DIFF
--- a/www/file/inner.js
+++ b/www/file/inner.js
@@ -86,7 +86,7 @@ define([
                 var $mt = $dlview.find('media-tag');
                 $mt.attr('src', src);
                 $mt.attr('data-crypto-key', 'cryptpad:'+cryptKey);
-                $mt.css('transform', 'scale(1)');
+                $mt.css('transform', 'scale(2)');
 
                 var rightsideDisplayed = false;
                 var metadataReceived = false;


davidbenque (approved):
The UI seems OK in the ctrl-e modal. Would need another quick review to check for unintended consequences.

davidbenque:
Taking a second look at this. There is no way to get the full title text in languages where the text overflows (aside from the accessibility label which seems correct). This doesn't seem like a good solution but adding tooltips to everything is also not great. Let's discuss the best way to solve this

PR #2131, opened 2025-12-21
fix(docs): Add proxy for sso and metrics plugins

### LINKED ISSUE #2042: sso/extensions.js not found
### Description

After upgrading to 2025.6.0 the `https://myinstance.example.com/sso/extensions.js` path could not be found anymore.
I've upgraded the nginx config as well and use the advanced one.

From what I can tell using `console.log` debugging the `sso` path is added to express and also points `Express.static` function from the sso plugin.

### Steps to reproduce

`cd lib/plugins`
`git clone https://github.com/cryptpad/sso.git`

Try to access `https://myinstance.example.com/checkup/`

### Relevant logs

```shell

```

### Expected behavior

Checkup page loads and sso extension definition is served

Linux/BSD/UNIX

_No response_

### Extensions

None

### Version

2025.6.0

promasu:
fixes #2042

### DIFF
--- a/docs/example-advanced.nginx.conf
+++ b/docs/example-advanced.nginx.conf
@@ -296,6 +296,21 @@ server {
         add_header Cross-Origin-Embedder-Policy require-corp;
     }
 
+    # Extensions are handeled via the server and have to be proxied
+    location ~ ^/(?:sso|metrics)/ {
+        proxy_pass http://localhost:3000;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header Host $host;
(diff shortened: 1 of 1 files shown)


ghost:
Hello @promasu,

> fixes #2042

Thanks a lot for your contribution! After discussion, we think it would be best to add this to each of the repositories:
- https://github.com/cryptpad/sso
- https://github.com/cryptpad/cryptpad-monitoring-plugin

Ideally, matching the way we do it for CryptPad, using a dedicated `/docs` folder.

PR #2133, opened 2026-01-02
Notifications for private messages

### LINKED ISSUE #1939: No notification for private messages
### Description

When sending a private message via the “contacts” app (Profile Picture > Contacts), the receiver doesn’t get notified of the arrival of a new message.

### Steps to reproduce

1. Log in;
2. Add a contact;
3. Open the “contacts” application (Profile Picture > Contacts);
4. Send a message to your contact;
5. They don’t know they received a message except if they open the “contacts” application.

### Expected behavior

The receiver should be informed, one way or another (Browser or via CryptPad’s bell) when they get a new direct message.

Note that browser’s notification got received if the “contacts” application is open in one of the browser’s tab.

Linux/BSD/UNIX

Mozilla Firefox, Chromium

### Extensions

Not relevant

### Version

2025.3.1

zuzanna-maria:
- users now get notifications when receiving private chat messages from other users (fixes #1939 )


yflory (changes requested):
The pull request in its current state should work but I think we could improve a few points:
- The notification is sent to the other user when you receive your own message from the server. It works but if you have the chat open in two tabs/windows on your account, you'll receive your message in both and you'll send the notification twice.
  - You should probably move the code to the [callback of the "send" function](https://github.com/cryptpad/cryptpad/blob/def39ed85a99ba6a8b3b5294cd80e7496daefa2e/www/common/messenger-ui.js#L399-L402), which is only called in the browser tab that sent the message
- You correctly dismiss the notifications if you already have one from the same user but you keep sending them for each message you write. It won't affect the UI but it adds unnecessary data to the server and the other user's mailbox. We need to discuss how to properly limit these notifications.
  - An easy way would be to stop sending notifications if you type multiple messages in a row (how many? and how to define "in a row"?).
  - A more complex system would be to stop sending notifications until the other user has acknowledged the first one. Basically, the other user would send you a mailbox message (`CHAT_NOTIFICATION_ACK`?) once they have closed the notification you sent them.
  - Another option, which could used in addition to one of the first two,  is to stop sending mailbox notifications if the other user is "online". In this case, if the recipient has the "contacts" page open but in the background, the client (`messenger-ui.js`) could trigger a notification from the "onMessage" function. And if the recipient has the contacts page open and visible, we don't need a notification at all.

yflory (approved):

Issue #2090, opened 2025-10-31
Can't load history snapshots in OnlyOffice documents


zuzanna-maria:
### Description

Cannot load snapshots created during browsing history of OO documents

### Steps to reproduce

1. Create OO document, make changes
2. Enter history mode and click on 'prev' to view the original (blank) version of the document
3. Create and save a snapshot of the original document state
4. Exit history view, returning to current version of the document
5. Attempt to open/load snapshot created while browsing history 
6. The document opens in its original state, not the state captured by the snapshot

### Relevant logs

```shell
n/a
```

### Expected behavior

Opening/loading a snapshot should show the document in the state it was at when snapshot was created, not its current state

### Extensions

n/a

### Version

2025.9.0

wginolas:
Sorry, I did not mean to close this issue

zuzanna-maria:
Closed with #2134

Issue #2089, opened 2025-10-31
Cannot share Sheet at a moment in history


zuzanna-maria:
### Description

When entering history mode in a sheet document, clicking 'prev' to return document to original state, and then 'share' to share the document at that moment in history, the link produced links to the current version of the document.

### Steps to reproduce

1. Create Sheet doc, make changes
2. Enter history mode and click 'prev' to see the doc in its original (blank) state
3. Click 'Share' and open the link
4. The linked document is in its current state 

### Relevant logs

```shell

```

### Expected behavior

The linked/shared document should be blank (like the original document, the history of which is being browsed and shared), not a copy of the current version of the original document 

### Extensions

n/a

### Version

2025.9.0

Issue #2039, opened 2025-08-27
Browsing history in documents with newlines doesn't work


zuzanna-maria:
### Description

When trying to revert to an earlier stage of a document which is a) freshly created and b) contains a new line/paragraph break, an error appears

### Steps to reproduce

1. Create new Document (this seemingly does not affect documents which have been previously created and are being re-opened, or which have been refreshed)
2. Enter text which includes a newline/paragraph break (using Enter)
3. Try to access a previous version by clicking on 'History' and then 'Previous version'
4. Either: an error message appears, or the document 'syncs' indefinitely

### Relevant logs

```shell
Uncaught TypeError: can't access property "Get_CurrentPage_Absolute", this.Content[nPos] is undefined
    controller_GetCurPage http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:22328
    GetCurPage http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:23495
    private_CheckCurPage http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:21633
    controller_RecalculateCurPos http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:22327
    RecalculateCurPos http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:23495
    RecalculateCurPos http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:21632
    CheckTargetUpdate http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:21628
    onTimerScroll_sync http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:28596
    OnResize http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all.js:28615
    Resize http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/word/sdk-all-min.js?ver=2025.6.0-1756198762047-1756303964290:3079
    onLayoutChanged http://localhost:3001/common/onlyoffice/dist/v8/web-apps/apps/documenteditor/main/app.js?ver=2025.6.0-1756198762047-1756303964290:8
[...]

Issue #2013, opened 2025-07-28
Can't restore Presentation history


zuzanna-maria:
### Description

Error when trying to restore Presentation history. 

### Steps to reproduce

1. Make changes to Presentation document
2. Enter history mode, go back to desired version
3. Click 'Restore'
4. Error modal appears
5. When page is refreshed, the correct (previously selected) version displays without issue

### Relevant logs

```shell
Uncaught TypeError: PosInfo.Class.Get_Id is not a function
    GetDocumentPositionBinary http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all-min.js?ver=2025.6.0-1753690791726-1753712438866:2482
    Get_DocumentPositionBinary http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all.js:2118
    Continue_FastCollaborativeEditing http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all.js:20751
    _autoSaveInner http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all-min.js?ver=2025.6.0-1753690791726-1753712438866:3112
    _autoSave http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all-min.js?ver=2025.6.0-1753690791726-1753712438866:2610
    onDocumentContentReady http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all-min.js?ver=2025.6.0-1753690791726-1753712438866:2601
    setInterval handler*baseEditorsApi.prototype.onDocumentContentReady http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all-min.js?ver=2025.6.0-1753690791726-1753712438866:2601
    _openDocumentEndCallback http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all-min.js?ver=2025.6.0-1753690791726-1753712438866:3271
    handleChanges http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all-min.js?ver=2025.6.0-1753690791726-1753712438866:1443
    _openDocumentEndCallback http://localhost:3001/common/onlyoffice/dist/v8/sdkjs/slide/sdk-all-min.js?ver=2025.6.0-1753690791726-1753712438866:3266
[...]

PR #2134, opened 2026-01-07
OnlyOffice history browsing

### LINKED ISSUE #2020: Cannot easily browse document history in OO
### Description

When trying to view the document history in Only Office apps, there is an option to jump all the way back to the creation of a document, and then gradually forward. There is no option at any point to jump back in increments unlike in other apps, which makes trying to find a particular version of a document difficult.

Image

OO document history

Image

Document history in non-OO apps

### Steps to reproduce

1. Create OO document and make changes
2. Click on history view and attempt to find a specific version of the document
3. It's not possible to gradually jump back to find a specific version -- only all the way back to creation and then gradually forward

### Relevant logs

```shell

```

### Expected behavior

It should be possible to navigate both forwards and backwards in small increments in all apps, including OO, when in history view

_No response_

_No response_

### Extensions

n/a

### Version

2025.6.0

zuzanna-maria:
- allows the browsing, sharing and creating snapshots of history in OnlyOffice documents (fixes #2020 , #2039 , #1772, #2013, #2089, #2090, #1421)

### DIFF
--- a/www/common/onlyoffice/history.js
+++ b/www/common/onlyoffice/history.js
@@ -37,6 +37,7 @@ define([
         var patch;
         var currentVersion;
         var forward;
+        var revertCheckpoint;
 
         // Get an array of the checkpoint IDs sorted their patch index
         var hashes = config.onlyoffice.hashes;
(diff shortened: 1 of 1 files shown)


yflory (changes requested):
In addition to the notes added to the code, I also have a few comments about the entire PR:
- I don't know what happened with this branch, but most of the commits are duplicated in the history. I'm not sure what would happen if we tried to merge this PR into the main code.
- I found an issue with the "version number". If you're on version `1.0`, press "back" to reach `0.x` and then "next", you end up on version `0.(x+1)` instead of `1.0`
- You should make sure the client is going to call `loadMoreOOHistory` only to load new data (try to never load the same data twice)

PR #2135, opened 2026-01-07
Fix empty file upload edge case (issue #2111)

### LINKED ISSUE #2111: Can not upload empty file on iOS
### Description

Can not upload empty file on iOS

The problem is the code is here: https://github.com/cryptpad/cryptpad/blob/main/src/common/common-util.js#L30
* `reader.result` is "data:" (without a comma) for an empty file
* Without a comma `res.indexOf(',')` does not work as expected

It might be a good idea to use [Uint8Array.toBase64()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64) instead.

https://github.com/cryptpad/cryptpad/issues/2112 is also related to this function and can probably be fixed together with this one.

### Steps to reproduce

1. Use an iOS device 
2. Upload an empty file to your drive.

Luckily browserstack provides an empty test file:
https://github.com/user-attachments/assets/acee7e30-cbb4-4b29-bcef-1a177f2c9e61

### Relevant logs

```shell

```

### Expected behavior

Empty files can be uploaded

Apple iOS

Apple Safari

### Extensions

.

### Version

2025.9.0

SaharYaz:
Fixes #2111

This PR fixes a bug when uploading an empty file. In some cases (reported on iOS), the upload data is data: without any content, and CryptPad crashes.

I added a safe check so empty uploads are handled correctly instead of failing.

Testing: I don’t have an iOS device, so I tested by simulating the data: input locally.

Let me know if you want changes or more tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -98,13 +98,74 @@
       }
     },
     "node_modules/@babel/helper-validator-identifier": {
-      "version": "7.27.1",
-      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.27.1.tgz",
-      "integrity": "sha512-D2hP9eA+Sqx1kBZgzxZh0y1trbuU+JoDkiEwqhQ36nodYqJwyEIhPSdMNd7lOm/4io72luTPWH20Yda0xOuUow==",
+      "version": "7.28.5",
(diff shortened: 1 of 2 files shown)


wginolas:
Hello @SaharYaz!

We fixed several issues in this part of CryptPad with https://github.com/cryptpad/cryptpad/pull/2181, which makes this PR obsolete. Sorry for the late response.

Still, thank you for your contribution and feel free to help in the future, if you want to!

PR #2144, opened 2026-01-19
moderator setting on by default

### LINKED ISSUE #2027: Help-desk answers must be anonymous by default
As a moderator on a CryptPad instance, I want the help-desk answers been sent anonymously by default.

Image

"Answer anonymously" should be enabled by default, answering not-anonymously should be opt-in.

achf01:
Closes #2027

Description:
Change from false to true the value of "privacy" in www/moderation/inner.js

### DIFF
--- a/www/moderation/inner.js
+++ b/www/moderation/inner.js
@@ -421,7 +421,7 @@ define([
         // Msg.support_privacyHint.support_privacyTitle
         sidebar.addCheckboxItem({
             key: 'privacy',
-            getState: () => false,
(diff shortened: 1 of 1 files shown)


davidbenque:
Hello @achf01, this only changes the value of the checkbox to "on" but the setting is not on by default. The moderator's username is still shown by default, and one has to un-check then check the box again to actually answer anonymously

achf01:
I'll link a new pull request with the tested fix.

PR #2165, opened 2026-01-29
Traefik docker compose added in community folder

cedricwalter:
* move caddy configs to community/caddy
* add traefik configs to community/traefik


cedricwalter:
I closed because I got instability, was working fine then suddenly no more, web socket timeout, will open when tested for more than a day  :-(

Issue #2175, opened 2026-02-04
Webmanifest integration


wallon-ines:
I would like to contribute to this project by adding a web manifest to use Cryptpad as a standalone / “native” app, for example, with Firefox PWA in Linux or Android in full-screen file, but what is the better way?

- Alter all `ogdata` constants in [build.js](https://github.com/cryptpad/cryptpad/blob/main/scripts/build.js#L159) ?
- Update [template.html](https://github.com/cryptpad/cryptpad/blob/main/customize.dist/src/template.html)  ?

Currently, I've just overridden files for my requirements, but if this can be useful to other people.

Example rendering : 

![Image](https://github.com/user-attachments/assets/91824c69-0697-43a0-9ea1-4b23dfec28dc)

<img width="3199" height="1799" alt="Image" src="https://github.com/user-attachments/assets/4c959f4e-06e8-4c3f-9d20-bd4d4d591c38" />

PR #2178, opened 2026-02-06
Feat #2175: Implement webmanifest

wallon-ines:
Add Web manifest
Issue: #2175

### DIFF
--- a/customize.dist/application.webmanifest
+++ b/customize.dist/application.webmanifest
@@ -1,0 +1,20 @@
+{
+    "name": "Cryptpad",
+    "short_name": "Cryptpad",
+    "description": "CryptPad: end-to-end encrypted collaboration suite",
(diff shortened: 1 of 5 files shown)


davidbenque:
Hello thanks for your contribution to CryptPad. We reviewed this for the current cycle but aren't exactly sure what to do with it since we don't fully understand the intent.

1. We understand this is towards a better integration of CryptPad as a Progressive Web App (PWA). However some gap will remain for CryptPad to fully be a PWA, for example we open a lot of new tabs all the time. Are there any benefits in merging what you propose without being a real PWA? will that not cause issues for users who might think they can "install" CryptPad and it not fully working?

2. We noticed the hard coded text in `customize.dist/application.webmanifest`, we have translations for these strings would there be a way to make use of them?

PR #2190, opened 2026-02-12
Set a maximum amount of responses for form

### LINKED ISSUE #2010: Maximum number of form responses
A user may wish to close a form automatically after a certain number of responses have been submitted.

This could be an optional value in form settings.

This feature may be used to prevent things like overbooking attendance. Without this option, form submissions must be frequently checked.

achf01:
Closes #2010. 
A way to set the maximum amount of responses is provided.
The limit can be set and unset. If the limit is reached the form is automatically closed and no other answers can be submitted. By increasing the limit the form will be reopened.


davidbenque (changes requested):
Hello and thanks for this contribution! I cannot comment on the code but already flagged an issue with translations that needs addressing.

davidbenque:
Hello sorry about the delay in replying but there are still issues with this PR. By default responses to a form aren't readable by participants, and if a participant cannot read responses it is still possible to submit past the limit in the way that this PR is implemented. Getting around this would require significant work that we are not sure is worth the feature proposed. Do you have use-cases in which this was needed?

PR #2219, opened 2026-02-27
Contacts page improvements

dariiing:
- improve layout for smaller screens
- update UI for no contacts
- update timestamp design
- add dropdown option
- fix username overflow
- update UI for history deletion
- improve accessibility
  -  add focus visible styles
  - fix tab order
  -  fix skip link
  - handle keyboard navigation
  - add aria-label

- new translation keys
   - update contacts_confirmRemoveHistory
   - remove contacts_info3 entirely


customize.dist/src/less2/include/messenger.less
```diff
@@ -112,13 +118,53 @@
...
+                            display: flex;
+                            align-items: center;
+                            background: none;
+                            border: none;
+                            color: inherit;
+                            cursor: pointer;
+                            padding: 2px;
+                            border-radius: @variables_radius;
+                            flex-shrink: 0;
+                            &:hover {
```
DianaXWiki, line 142: I’d switch this hover color to a theme background token so it stays consistent across light and dark themes. Right now it doesn’t match the chat send button hover color; using @msg-bg-color-dark (and var(--msg-bg-color-dark)) seems to align behavior

customize.dist/src/less2/include/messenger.less
```diff
@@ -257,29 +317,41 @@
...
             .hover () {
                 cursor: pointer;
-                height: 100%;
-                line-height: 30px;
                 padding: 10px;
+                border-radius: @variables_radius;
                 &:hover {
-                    background-color: @msg-bg-color-darker;
-                    background-color: var(--msg-color-darker);
+                    background-color: rgba(0, 0, 0, 0.2);
```
DianaXWiki: Here the same, to be reviewed again, or if not consistent with the desired styling to introduce a new variable with both colors for light and dark

www/common/messenger-ui.js
```diff
@@ -238,20 +239,66 @@ define([
...
+                showBanner($chat, id);
+            }
+            common.setAttribute(['general', 'chat', 'cleared', id], true);
         };
         markup.chatbox = function (id, data, curvePublic) {
             var moreHistory = h('span', {
                 class: 'cp-app-contacts-more-history',
+                tabindex: '0',
+                role: 'button',
+                'aria-label': Messages.contacts_fetchHistory
```
DianaXWiki: When navigating with keyboard, the “more history” control behaves like it has duplicate focus (it takes two tab actions and the tooltip appears on the second). This seems to come from having a focusable wrapper (span with tabindex/role="button") plus an inner icon with title

www/common/messenger-ui.js
```diff
@@ -308,34 +355,58 @@ define([
             });
 
             var removeHistory = h('span', {
-                'class': 'cp-app-contacts-remove-history'
+                'class': 'cp-app-contacts-remove-history',
+                'tabindex': '0',
+                'role': 'button',
+                'aria-label': Messages.contacts_removeHistoryTitle
```
DianaXWiki: Same duplicate focus for the removeHistory button

www/common/messenger-ui.js
```diff
@@ -456,6 +528,29 @@ define([
...
         };
 
+        var updateInfoMessage = function () {
+            var chats = Object.keys(state.channels).length;
+            var $info = $messages.find('.cp-app-contacts-info');
+            if (chats === 0) {
+                $info.html([
+                    h('h2', Messages.contacts_noFriends),
+                    h('ul', [
+                        h('li', Messages.share_noContactsNotLoggedIn),
```
DianaXWiki: This message is also displayed for logged in users which can be confusing

www/common/messenger-ui.js
```diff
@@ -543,85 +639,140 @@ define([
...
+            var friendData = room.isFriendChat ? userlist[0] : {};
+
+            var dropdownOptions = [];
+
+            var removeOption = {
+                tag: 'a',
+                content: [Icons.get('unfriend'), h('span', Messages.contacts_remove)],
+                action: function () {
+                    var channel = state.channels[id];
+                    if (!channel.isFriendChat) { return; }
```
DianaXWiki: Also verify that channel exists and `return true` for consistency

www/common/messenger-ui.js
```diff
@@ -543,85 +639,140 @@ define([
...
+                }
+                $dropdown.setOptions(opts);
+                $dropdown.find('.cp-dropdown-content').hide();
+            };
+
+            var muteOption = {
+                tag: 'a',
+                content: [Icons.get('mute'), h('span', Messages.contacts_mute || 'Mute')],
+                action: function () {
+                    var channel = state.channels[id];
```
DianaXWiki, line 682: Check for channel

www/common/messenger-ui.js
```diff
@@ -543,85 +639,140 @@ define([
...
+                    }
+                    opts.push(removeOption);
+                }
+                $dropdown.setOptions(opts);
+                $dropdown.find('.cp-dropdown-content').hide();
+            };
+
+            var muteOption = {
+                tag: 'a',
+                content: [Icons.get('mute'), h('span', Messages.contacts_mute || 'Mute')],
```
DianaXWiki: I don't think we need the plain text fallback

www/common/messenger-ui.js
```diff
@@ -543,85 +639,140 @@ define([
...
+                    if (!channel.isFriendChat) { return true; }
+                    var curvePublic = channel.curvePublic;
+                    unmuteUser(curvePublic);
+                    rebuildDropdown(false);
+                    return true;
+                }
+            };
+
+            if (room.isFriendChat) {
+                dropdownOptions.push(isMuted ? unmuteOption : muteOption);
```
DianaXWiki, line 705: This dropdown unmute option seems to be not synced with the one in the `Manage Muted` section. If you unmute from that pop-up and open again the dropdown you can still see the Unmute option even if the user is already unmuted

www/common/messenger-ui.js
```diff
@@ -543,85 +639,140 @@ define([
...
+                    top: rect.bottom + 'px',
+                    left: left + 'px',
+                });
+            });
+
             var status = h('span.cp-app-contacts-status', {
                 title: Messages.contacts_online
             });
+            var mute = h('span.cp-app-contacts-mute-indicator', {
+                title: Messages.contacts_mute,
```
DianaXWiki: The tooltip on cp-app-contacts-mute-indicator currently says “Mute”, but this icon represents current state, not an action. Could we use a dedicated key like contacts_muted (“Muted”) for clarity?

www/common/messenger-ui.js
```diff
@@ -802,6 +953,10 @@ define([
                 });
                 normalizeLabels($messagebox);
 
+                common.getAttribute(['general', 'chat', 'cleared', id], function (err, val) {
+                    if (val) { showBanner($(chatbox), id); }
+                });
+
```
DianaXWiki: There are some added empty lines like this in the code that should be removed

www/common/messenger-ui.js
```diff
@@ -76,11 +75,11 @@ define([
                 h('div.cp-app-contacts-category-content')
             ]),
             h('div.cp-app-contacts-friends.cp-app-contacts-category', [
+                h('div.cp-app-contacts-category-content.cp-contacts-friends'),
                 h('button.btn.btn-default.cp-app-contacts-muted-button', {tabindex:0},[
                     Icons.get('mute'),
                     Messages.contacts_manageMuted
```
DianaXWiki, line 82: Add aria-label here as well

www/common/messenger-ui.js
DianaXWiki: Also not related to this issue, but long user's names should also be handled in the `Managed Muted` pop-up

customize.dist/src/less2/include/messenger.less
DianaXWiki: And keep consistency with the hover styling of the send message button

DianaXWiki (changes requested):

www/common/messenger-ui.js
```diff
@@ -238,20 +239,66 @@ define([
...
+            var $chat = $(getChat(id));
+            if (state.channels && state.channels[id]) {
+                state.channels[id].messages = [];
+            }
+            if ($chat.length) {
+                $chat.find('.cp-app-contacts-messages').html('');
+                $chat.find('.cp-app-contacts-message').remove();
+                showBanner($chat, id);
+            }
+            common.setAttribute(['general', 'chat', 'cleared', id], true);
```
yflory: We shouldn't use setAttribute here. It would create data in the account that is never removed and it would grow for each contact.
Also, I don't really understand the goal of this value. If you clear a chat:
1. you'll keep seeing the banner "chat history has been deleted" until you close it, even if you type new content in the chat
2. the other user won't see this message

I think a better change would be to have this message as a temporary message in your UI, but also send a specific type of message to the chat to warn the other user that the history has been deleted.

yflory (changes requested):

www/common/messenger-ui.js
yflory: I'm not sure we should `return true` in this case (which has a different meaning than just `return`, so not really a consistency issue). This is clearly an error and something that shouldn't happen so maybe we could add an error message with `UI.warn`
```js
UI.warn(Messages.error); // or a custom error message
return;
```

yflory (approved):

PR #2226, opened 2026-03-03
Enforce immediate access-list lockout and prevent stale content visibility on refresh

### LINKED ISSUE #2191: Enabling access list does not kick existing sessions
### Description

If you share a document via share link, then enable the access list after the guest has opened the page, the stale contents are still visible.

Tested with rich text document.

### Steps to reproduce

1. Share document (with no access list) via share link.
2. Open that share link in incognito window. View contents as guest.
3. Enable access list (guest is not on acl).
4. An auth popup appears, but the stale contents are still visible in the background as readonly. Upon refresh the user can see the stale contents.

### Relevant logs

```shell

```

### Expected behavior

Guest is immediately kicked and can see nothing, even after refresh.

Of course they can still save the data manually before this point, record their screen, etc, but this basic protection would help in the overwhelming majority of normal cases where access is revoked.

Linux/BSD/UNIX

Other

### Extensions

Using brave on cryptpad.fr instance.

Extensions: uBlock origin, bitwarden, but I suspect this is server-side behavior.

### Version

2026.1.0

DianaXWiki:
This PR aims to fix #2191:

- [x] Ensure users removed from access (or not included in the allow list) or guests are immediately blocked 
- [x] Prevent stale document content from being rendered on guest read-only refresh
- [x] This PR should fix the issue for both documents and shared folders (for both personal and team drives)


www/drive/inner.js
```diff
@@ -57,6 +70,11 @@ define([
                     sharedFolder: fId
                 }, waitFor(function (err, newObj) {
                     if (!APP.loggedIn && APP.newSharedFolder) {
+                        if (err === 'ERESTRICTED' || (newObj && newObj.restricted)) {
```
yflory: `err` only indicates communication errors between the iframe and its parent and so it can never be "ERESTRICTED" here (which comes from the server).

www/drive/inner.js
```diff
@@ -57,6 +70,11 @@ define([
...
+                            lockoutAnonSharedFolder(common);
+                            lockedOut = true;
```
yflory: You don't need this `lockedOut` system here. You can use `waitFor.abort();` to force the code to stop here and ignore the following nThen blocks (line 121 in this case).

yflory (changes requested):
I've added two small comments to slightly improve the code quality but the current version should work.

yflory (approved):

PR #2238, opened 2026-03-13
Diagram initialized in read-only mode until document is ready

### LINKED ISSUE #1407: New Diagram initialization has a race condition
### Description

The diagram app is initialized with an "editable" state, so we can start editing even before a document is "ready". It should be initialized in read-only mode

### Steps to reproduce

.

### Relevant logs

_No response_

### Expected behavior

.

_No response_

_No response_

### Extensions

.

### Version

5.7.0

zuzanna-maria:
- fixes #1407 (diagram initialized in read-only mode, switches once document is ready)


wginolas (approved):
👍

PR #2243, opened 2026-03-17
Information about share links on support page

### LINKED ISSUE #2241: Include notice re: Share links on Support page
### Description

The current Support page text doesn't mention the best way to share documents which are affected by a bug/issue which the user is reporting. Consequently, users most often share document IDs, unaware that admins/moderators can't use them to open the affected document to diagnose a problem.

### Steps to reproduce

1. Have access to an instance's support page
2. When users report issues with a document, they don't include a share link (instead send a document ID, or nothing at all)
3. Can't view the document in question to diagnose the issue (e.g. via console logs)
4. Have to ask user to send a share link, leading to unnecessary back-and-forth

### Relevant logs

```shell

```

### Expected behavior

The text on the Support page should be updated to advise users on which document information to share in order for the support team to localize the document 

_No response_

_No response_

### Extensions

n/a

### Version

2026.2.0

zuzanna-maria:
- fixes #2241 (a notice requesting users include a share link to documents, in addition to document identifiers, in support tickets, is now included on the Support page)

### DIFF
--- a/customize.dist/messages.js
+++ b/customize.dist/messages.js
@@ -134,6 +134,9 @@ define(req, function(AppConfig, Default, Language) {
         }
     };
 
+    // Change to existing key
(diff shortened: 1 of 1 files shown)


davidbenque:
After discussion we will try to improve support messaging a different way (adding a category and possibly checking for "safe links")

PR #2250, opened 2026-03-19
Pad: Include media-tag CSS in print preview so images sized correctly

### LINKED ISSUE #2037: In print view for rich text document, image size is set in the wrong element
### Description

In the webpage that's opened using the print button on a rich text document, the CSS width and height is set on the media-tag element, but instead it should be set on the img tag. This results in the size being ignored.

### Steps to reproduce

1. create rich text document with image
2. click file
3. click print
4. see that the image is too big
5. use browser "inspect element" button on the image
6. try to edit the width property, and see that there's no effect
7. add the width property to the inner img element, and see that there is an effect

### Relevant logs

```shell

```

### Expected behavior

size should be handled correctly

Linux/BSD/UNIX

Mozilla Firefox

### Extensions

a bunch of extensions

### Version

2025.6.0

WebCoder49:
Fixes #2037 

This patch does not change the HTML of the documents produced by the editor, so should be backwards-compatible with documents containing images previously sized using the dialog.

The fundamental requirement is to make a piece of CSS exist in both the editor and print preview window. Here's a method using a regular expression and the `contentPreview` event to directly insert the CSS into the HTML, since it's what CKEditor's own MathJax plugin does (for a JavaScript import).

This isn't disgusting, in my opinion. Nevertheless, there may be a better, cleaner way, but `CKEDITOR.addCss` is not it. Please let me know if there is one you know of. I don't know whether CKEditor 5 does this in a cleaner way, but it would be easier to suggest a cleaner approach upstream in CKEditor's repository if Cryptpad used CKEditor 5. That might be a good (but difficult) idea for the long-term because the CKEditor 4 we use is no longer supported.

### DIFF
--- a/www/pad/mediatag-plugin.js
+++ b/www/pad/mediatag-plugin.js
@@ -4,31 +4,32 @@
 
 ( function() {
 
+    const mediaTagCss = (
+        'media-tag{' +
+            'display:inline-block;' +
+            'border-style: solid;' +
+            'border-color: black;' +
+            'border-width: 0;' +
+        '}' +
+        'media-tag.selected{' +
+            'border: 1px solid black;' +
+        '}' +
+        'media-tag iframe{' +
+            'border: 6px solid #eee;' +
+        '}' +
+        'media-tag img{' +
+            'vertical-align: top;' +
+        '}' +
+        'media-tag *{' +
+            'width:100%; height:100%;' +
+        '}');
+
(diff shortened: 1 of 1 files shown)


WebCoder49:
Sorry for all the small commits - I'm now squashing them into one.

yflory (approved):

PR #2296, opened 2026-05-15
Fix dependencies with missing resolved in lockfile

eljamm:
This change regenerates the lockfile to fix the missing `resolved` fields:

```shellSession
rm package-lock.json & npm install --package-lock-only --ignore-scripts
```

Which causes issues while building Cryptpad in reproducible environments, for example when [updating in Nixpkgs](https://github.com/NixOS/nixpkgs/pull/511890):

```shellSession
warning: 313 out of 483 packages (64.8%) are missing 'resolved' URLs and will not be cached.
warning: Packages without 'resolved' URLs:
warning:   - jwa
warning:   - safer-buffer
warning:   - methods
warning:   - minitask
warning:   - buffer-from
warning:   - looper
warning:   - @node-saml/node-saml
warning:   - jszip/node_modules/safe-buffer
warning:   - accepts
warning:   - gluejs/node_modules/resolve
warning:   ... and 303 more
warning: More than 50% of packages are missing 'resolved' URLs. This may indicate an issue with the lockfile.
warning: This is a known issue with some npm versions. See: https://github.com/npm/cli/issues/6301
warning: Consider regenerating upstream's lockfile with: npm install --package-lock-only (sending an upstream PR is best)
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -84,11 +84,13 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.27.1",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
+      "integrity": "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


eljamm:
Resolved merge conflict.

martinetd:
Hi @davidbenque , thanks for the review & fixing the branch to staging last month

@yflory , would you be able to have a second look so this could get unstuck? I assume 2026.6 is just around the corner and it'd be great to fix this so we can update the nixpkgs package :)
(sorry for the direct ping, you merged my PRs in the past)

Chouhartem:
Hello,

Sorry for the late reply but we will integrate these changes in the next release :)

PR #2308, opened 2026-06-04
Fixed error propagation in integration mode

Max-7:
In integration mode, I've noticed some errors were detected but `onError` was not called and the "parent" app was not informed.

`sframe-common-outer` listens to `Q_INTEGRATION_ERROR`, which doesn't seem to exist (a grep doesn't return anything) instead of `EV_INTEGRATION_ERROR`.

### DIFF
--- a/www/common/sframe-common-outer.js
+++ b/www/common/sframe-common-outer.js
@@ -2266,7 +2266,7 @@ define([
                 sframeChan.on('Q_INTEGRATION_USERLIST_CHANGE', function (obj, cb) {
                     cfg?.integrationUtils?.onUserlistChange?.(obj, cb);
                 });
-                sframeChan.on('Q_INTEGRATION_ERROR', function (obj) {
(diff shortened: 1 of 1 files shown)


Chouhartem (approved):
Thanks for catching this, looks good to me.

wginolas (approved):

wginolas:
This PR fixes https://github.com/cryptpad/cryptpad/issues/2309

Max-7:
> This PR fixes #2309

Are you sure it does? The problem (as I understand it) with #2309 is that when the conversion fails, the error is displayed but not really processed, and `""` is returned.
https://github.com/cryptpad/cryptpad/blob/9808cf25c1091d6cf532df13bf5a70ba332f8d4d/www/common/outer/x2t.js#L217
It then leads to an empty document being converted to a valid empty docx. I think that properly fixing #2309 requires interrupting the conversion if it fails and calling the callback with null data, which in turns get propagated properly with this PR.

yflory (approved):

PR #2318, opened 2026-06-12
Fix OnlyOffice spellcheck

### LINKED ISSUE #2142: Spellcheck not working in OnlyOffice
### Description

Spellcheck is not working in OnlyOffice documents despite being selected in OnlyOffice settings - the only place it works is the initial slide of Presentation documents. 

### Steps to reproduce

1. Create OnlyOffice document of any type
2. Ensure spellcheck is enabled in document settings
3. If sheet/doc, spellcheck will not work at all. If presentation, it will only work on the first slide

### Relevant logs

```shell

```

### Expected behavior

Spellcheck works for all text in all document types if enabled (and doesn't work if disabled).

_No response_

_No response_

### Extensions

n/a

### Version

2025.9.0

zuzanna-maria:
- fixes #2142 (OnlyOffice apps load en-us spellcheck for English language users on each document load)

### DIFF
--- a/www/common/onlyoffice/inner.js
+++ b/www/common/onlyoffice/inner.js
@@ -2129,7 +2129,8 @@ define([
                     });
                 }
 
-                var defaultLanguageCode = String(getEditor().asc_getDefaultLanguage());
+                //Ensure spellcheck set correctly (works only with localized language variants)
+                var defaultLanguageCode = getEditor().asc_getDefaultLanguage();
 
                 const langToLocale = {
                     "en" : "en-us",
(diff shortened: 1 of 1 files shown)


www/common/onlyoffice/inner.js
```diff
@@ -2129,12 +2129,11 @@ define([
...
-                    var w = getWindow();
-                    if (lang === "fr") { lang = 'fr-fr'; }
-                    var l = w.Common.util.LanguageInfo.getLocalLanguageCode(lang);
-                    getEditor().asc_setDefaultLanguage(l);
-                }
+                var w = getWindow();
+                if (lang === "en") { lang = "en-us"; }
+                if (lang === "fr") { lang = 'fr-fr'; }
+                var l = w.Common.util.LanguageInfo.getLocalLanguageCode(lang);
+                getEditor().asc_setDefaultLanguage(l);
```
wginolas: I think we only should do this for new documents. Otherwise, a multi-language team would constantly change the language of a shared doc.

wginolas (changes requested):

wginolas (approved):
Oh, I did not know we would handle three different types of language codes. It looks good to me.

yflory (approved):

yflory (approved):

PR #2324, opened 2026-06-26
Add option to remove team avatar

DianaXWiki:
This PR aims to improve teams UI by adding a button to remove the set picture for the team avatar for administrators.

### DIFF
--- a/www/teams/inner.js
+++ b/www/teams/inner.js
@@ -683,12 +683,12 @@ define([
         APP.module.execCommand('GET_TEAM_METADATA', {
             teamId: APP.team
         }, (obj) => {
-            if (obj && obj.error) { return; }
+            if (obj?.error) { return; }
             const $headerAvatar = APP.$leftside.find('.cp-team-cat-header .cp-avatar');
             if ($headerAvatar.length) {
                 const $name = $headerAvatar.find('.cp-sidebarlayout-category-name').detach();
                 $headerAvatar.empty();
-                common.displayAvatar($headerAvatar, obj.avatar, obj.name);
+                common.displayAvatar($headerAvatar, obj?.avatar, obj?.name);
                 if ($name.length) { $headerAvatar.append($name); }
             }
             APP.refreshTeamAvatarBlock(obj);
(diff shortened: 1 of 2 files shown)


dariiing (changes requested):
missing a right margin here:
image

www/teams/inner.js
```diff
@@ -1172,8 +1188,41 @@ define([
...
+                        }, function () {
+                            refreshTeamMetadataDisplay(common);
+                        });
+                    });
+                });
+            });
+        };
+        var refreshAvatar = function (obj) {
+            displayAvatar(obj);
+        };
```
Chouhartem: Why redefining a function here and not name it directly `refreshAvatar`?

+ On the previous code, the function was named `redrawAvatar` (for non-team avatar): https://github.com/cryptpad/cryptpad/blob/main/www/settings/inner.js#L1873

dariiing (approved):

www/teams/inner.js
```diff
@@ -1171,50 +1187,71 @@ define([
...
+                    }, (meta) => {
+                        if (meta?.error) { return void UI.warn(Messages.error); }
+                        meta.avatar = '';
+                        APP.module.execCommand('SET_TEAM_METADATA', {
+                            teamId: APP.team,
+                            metadata: meta
+                        }, () => {
+                            refreshTeamMetadataDisplay(common);
+                        });
+                    });
```
yflory: This code is duplicate from the callback of `addAvatar` below, except that the avatar value is set to empty string instead of a URL. It would be better to move it to a new `setAvatar(url)` function which would be called in these 2 cases (`setAvatar('');` and `setAvatar(upload.url);`)

www/teams/inner.js
```diff
@@ -678,6 +678,22 @@ define([
...
+        }, (obj) => {
+            if (obj?.error) { return; }
+            const $headerAvatar = APP.$leftside.find('.cp-team-cat-header .cp-avatar');
+            if ($headerAvatar.length) {
+                const $name = $headerAvatar.find('.cp-sidebarlayout-category-name').detach();
+                $headerAvatar.empty();
+                common.displayAvatar($headerAvatar, obj?.avatar, obj?.name);
+                if ($name.length) { $headerAvatar.append($name); }
+            }
+            APP.refreshTeamAvatarBlock(obj);
```
yflory: You should make sure that `APP.refreshTeamAvatarBlock` exists and is a function here. You could end up with a TypeError otherwise.
There is also a small risk of race condition where refreshTeamAvatarBlock may represent a different team that the one you're currently viewing. You could delete this value when switching to a different team (`delete APP.refreshTeamAvatarBlock;`) to avoid this issue.

yflory (changes requested):

yflory (approved):

PR #2329, opened 2026-07-09
Kanban accessibility fixes

### LINKED ISSUE #1810: Lack of tooltip inside for Kanban
This button doesn't have a tooltip and it might be confusing for some users, not knowing what it does.

Image

dariiing:
This PR fixes #1810 and adds labels to the color picker on the kanban

### DIFF
--- a/customize.dist/messages.js
+++ b/customize.dist/messages.js
@@ -134,6 +134,8 @@ define(req, function(AppConfig, Default, Language) {
         }
     };
 
+    Messages.kanban_quickView = "Minimize card content"; // XXX
+    Messages.kanban_fullView = "Show card content";
     return Messages;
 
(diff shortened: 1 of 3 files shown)


DianaXWiki (commented):
The fixes look good and work as expected. Just a few notes on kanban accessibility in general:
- **Modal focus**: Opening the edit modal does not move focus into the dialog, but rather you have to tab through many page elements before reaching it. Expected: focus should land on the title field (or first focusable control) on open
- **Esc key in tags field**: Pressing Esc in the tags input closes the modal instead of moving focus to the next field (colors)

DianaXWiki (approved):

yflory (approved):

PR #2332, opened 2026-07-09
Prevent .webp insertion in OnlyOffice

### LINKED ISSUE #2044: Prevent users from using .webp images in OnlyOffice documents
As long we can not export documents with `.webp` images, we should not allow users to add these images to a document.

See also: https://github.com/cryptpad/cryptpad/issues/1504

zuzanna-maria:
- fixes #2044 (users prevented from inserting .webp images into OnlyOffice documents)

### DIFF
--- a/www/secureiframe/inner.js
+++ b/www/secureiframe/inner.js
@@ -187,6 +187,7 @@ define([
                 if (to) { window.clearTimeout(to); }
                 to = window.setTimeout(updateContainer, 300);
             });
+
             // If file, display the upload button
             if (types.indexOf('file') !== -1) {
                 var f = (filters && filters.filter) || {};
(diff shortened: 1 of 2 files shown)


www/secureiframe/inner.js
```diff
@@ -197,6 +197,9 @@ define([
                         if (/^[a-z]+\/$/.test(val)) {
                             val += '*';
                         }
+                        if (val === 'image/*' && (priv.app === 'doc' || 'sheet' || 'presentation')) {
+                            val = ['image/png', 'image/svg+xml', 'image/jpeg'];
```
Chouhartem: Hello,

So far svg images don’t appear to be working in office apps. We can take advantage of this PR to block them as well.

yflory (changes requested):
We already have code to select which file types can be added to a document. For office documents, you can do it [here](https://github.com/cryptpad/cryptpad/blob/9808cf25c1091d6cf532df13bf5a70ba332f8d4d/www/common/onlyoffice/inner.js#L2457)

yflory (approved):

PR #2335, opened 2026-07-13
Remove option to add file already in team drive to team

### LINKED ISSUE #1948: Sharing a file which is in a team drive offers an option to add it to the team drive
### Description

From within a document the share button opens a dialog which includes the option to "Add to team drive".
This is confusing when the document is already within that team drive.

### Steps to reproduce

1. Create team
2. Create document in that team
3. Share document

### Relevant logs

```shell

```

### Expected behavior

It should indicate in some way that the document is already part of that team

Microsoft Windows

Mozilla Firefox

### Extensions

Add-ons Search Detection	extension	2.0.0	true	app-builtin-addons	addons-search-detection@mozilla.com
Awesome Emoji Picker	extension	2.0	true	app-profile	awesome-emoji-picker@rugk.github.io
Facebook Container	extension	2.3.12	true	app-profile	@contain-facebook
Form Autofill	extension	1.0.1	true	app-builtin-addons	formautofill@mozilla.org
Kagi Search for Firefox	extension	0.7.6	true	app-profile	search@kagi.com
Picture-In-Picture	extension	1.0.0	true	app-builtin-addons	pictureinpicture@mozilla.org
PubPeer	extension	1.6.2	true	app-profile	jid1-rU2mNakSg7IiSQ@jetpack
Scite	extension	1.36.2	true	app-profile	{8bc17535-7be0-4000-825b-f87c7b45a65a}
Unpaywall	extension	3.98	true	app-profile	{f209234a-76f0-4735-9920-eb62507a54cd}
Web Compatibility Interventions	extension	140.7.20250522.151919	true	app-system-addons	webcompat@mozilla.org
Zotero Connector	extension	5.0.151	true	app-profile	zotero@chnm.gmu.edu
System theme — auto	theme	1.4.1	true	app-builtin	default-theme@mozilla.org

### Version

2025.3.1

zuzanna-maria:
- fixes #1948 (can no longer share a file with a team if already added to team drive)


www/common/inner/share.js
```diff
@@ -88,6 +88,23 @@ define([
                 large: true,
                 data: teams
             }, refreshButtons);
+
+            common.getFilesList(['file', 'pad'], function (err, list) {
```
yflory: Getting all the documents stored in a team from the worker to this secure iframe and only then checking if the current one is stored seems inefficient. It would probably be better to add a new command which returns the list of teams which contain this pad.

www/common/inner/share.js
```diff
@@ -88,6 +88,23 @@ define([
...
+                teamsList.icons.forEach(function(icon) {
+                    var inTeam = Object.values(list).some(file =>
+                        file.teamId === icon.getAttribute('data-teamid') &&
+                        file.channel === config.channel
+                    );
+                    if (inTeam) {
+                        $(icon).addClass('cp-disabled');
```
yflory, line 101: This code is asynchronous so here you're going to "disable" this team only after the worker returns the value. If for some reason the worker can't return the value or is too slow to return it, the user would still be able to click on this team.
It would be better to compute the list of teams already containing this pad before creating the "user grid".

yflory (changes requested):

www/common/cryptpad-common.js
```diff
@@ -1246,6 +1246,11 @@ define([
             cb(void 0, list);
         });
     };
+    common.getPadTeams = function (query, cb) {
+        postMessage("GET_PAD_TEAMS", query, function (teams) {
+            cb(void 0, teams);
```
yflory: Here you should `cb(teams);` directly, there is no need to create an always "undefined" argument.
In case of error, `teams` will be an object wiht an "error" value (`teams.error`);

www/secureiframe/main.js
```diff
@@ -162,6 +162,16 @@ define([
...
                 });
 
+
+                sframeChan.on('Q_GET_PAD_TEAMS', function (types, cb) {
+                    Cryptpad.getPadTeams(types, function (err, data) {
+                        cb({
+                            error: err,
+                            data: data
+                        });
+                    });
```
yflory: `err` will always be undefined here (see the comment from cryptpad-common.js) so it can be removed. Additionally, `data` is either an object containing an error or an array with the expected result, so you can also use
```js
                    Cryptpad.getPadTeams(types, cb);
```

www/common/sframe-common.js
```diff
@@ -689,6 +689,12 @@ define([
         });
     };
 
+    funcs.getPadTeams = function (query, cb) {
+        ctx.sframeChan.query('Q_GET_PAD_TEAMS', query, function (err, data) {
+            cb(err || data.error, data.data);
```
yflory: Following the changes suggested in sframe-common-outer.js, this should be:
```js
cb(err || data?.error, data);

File: www/common/inner/share.js [hunk 6]
```diff
@@ -89,12 +89,14 @@ define([
                 data: teams
             }, refreshButtons);
 
-            common.getFilesList(['file', 'pad'], function (err, list) {
+
+            common.getPadTeams({channel: config.channel}, function (err, teamIds) {
+                if (err) { return void console.error(err); }
```
yflory: In case of error, we probably shouldn't `return` here and only log the error. That way this PR won't break the existing "feature" which already removes the current team (if you're in a team drive).

yflory (changes requested):

www/secureiframe/main.js
```diff
@@ -162,6 +162,11 @@ define([
                     });
                 });
 
+
+                sframeChan.on('Q_GET_PAD_TEAMS', function (types, cb) {
```
yflory: You should move this handler to sframe-common-outer.js inside the `addCommonRpc` function because we also need to call it from the drive.
"addCommonRpc" defines the list of commands that can be received both by the secure iframe and by the sandboxed iframe.

yflory (changes requested):

yflory (approved):

Issue #2214, opened 2026-02-25
Incorrect formatting for numerical values in Form answers imported to Sheet


zuzanna-maria:
### Description

From a support ticket: when numerical Form answers are imported into Sheet, they cannot be summed with a formula, as they are an incorrect format (and cannot be reformatted using highlight>right click>format cells)

### Steps to reproduce

1. Make a Form with a (required) question type 'Text' and Text type 'Number'. 
2. Answer the Form several times so you have several numbers.
3. Go to the Responses and click on the button 'Export to Sheet'
4. Add a formula to sum the cells with the numbers.
5. The result of the sum is zero.
6. When selecting the cells with the numbers, the status bar only shows the count.

### Relevant logs

```shell

```

### Expected behavior

- The result of the sum is the count of the numbers in summed cells.
- When selecting the cells with the numbers, the status bar shows the average, count, min, max and sum.

### Extensions

n/a

### Version

2026.2.0

josep11:
+1 here.

It does not do anything at all. No way to convert to number.

I attach a sample of an exported file, for if someone wants to troubleshoot faster. I imported it to CryptPad again and the behaviour is the same.

[Export-sample-formulari-comanda.ods](https://github.com/user-attachments/files/29630929/Export-sample-formulari-comanda.ods)

josep11:
Any news on this?

josep11:
Okay, I could diagnose what is the issue here.

### Steps
1. I downloaded the exported form as .ods
2. Uncompress it with: `unzip -o -q form_exported.ods -d form_extracted`
3. Cd into that directory and print the contents: `xmllint --format content.xml`

```
<table:table-cell office:value-type="string">
      <text:p>2</text:p>
</table:table-cell>
```

We can consistently observe that the issue is that it does not discriminate at the point of exporting and every cell is marked with `office:value-type="string"`. This is not what we want, it should be `office:value-type="float"`.

### Conclusion
Confirmed that the issue is at export time

josep11:
I submitted a PR here: #2340

PR #2340, opened 2026-07-17
Fix issue #2214 incorrect formatting of numerical values in form answers

josep11:
Implement fix for issue #2214. Created a complimentary array in order not to mess with the other export use-cases (CSV and JSON).

- Unanswered/empty number questions stay as '' (not coerced to 0 or NaN).
- Types with a custom exportCSV (radio, checkbox, sort, poll) are left untouched -- they're categorical/textual by nature and out of scope for this issue.
- No changes to CSV or JSON export use-cases.


zuzanna-maria (changes requested):
Hello @josep11 , thank you for this contribution. The PR works well to fix the issue. However, we noticed you added some unit tests for this PR in a format different than is used in our codebase, as well as a new test runner. We feel that this is outside of the scope of this PR, and so we would like to ask that you remove the new files and the new package, and keep only changes to `export.js`. Thanks!

josep11:
> Hello @josep11 , thank you for this contribution. The PR works well to fix the issue. However, we noticed you added some unit tests for this PR in a format different than is used in our codebase, as well as a new test runner. We feel that this is outside of the scope of this PR, and so we would like to ask that you remove the new files and the new package, and keep only changes to `export.js`. Thanks!

Okay, thanks!
What files, exactly? Not the unit test for export.js, right?

josep11:
@zuzanna-maria, I've stripped the 3 changes related to testing the fix onto PR https://github.com/cryptpad/cryptpad/pull/2378. So, this one now contains only the fix.

Let's debate there about the unit test, which I think needs to be part of the CI pipeline for early detection of regressions.

josep11:
Hi @zuzanna-maria, let me know on the tests.
Btw, I cannot change the destination branch to staging.

PR #2349, opened 2026-07-30
Improve sharing forms UX

DianaXWiki:
This PR introduces improvements for better handling of share forms UI:

- [x] The creation screen now prompts the user to choose a place to save their new document (CryptDrive, a team, or don’t store). “Don’t store” is an explicit choice rather than easy to miss.
- [x] For guests users there have been several modals introduced explaining the steps of sharing forms and different links functionality:
      - For key actions such as "Copy public link" or "Preview" the guest user is always announced about the importance of the author link and the necessity of saving it.

### DIFF
--- a/www/common/common-ui-elements.js
+++ b/www/common/common-ui-elements.js
@@ -3149,17 +3149,11 @@ define([
             var $template = $creation.find('.cp-creation-template-selected');
             var templateId = $template.data('id') || undefined;
             var templateContent = $template.data('content') || undefined;
-            // Team / drive storage
-            // -1 ==> own CryptDrive; team id ==> team drive; none ==> nowhere
+            // Team
             var team;
-            if (teamValue === '-1') {
-                team = { id: -1 };
-            } else if (teamValue && teamValue !== 'none') {
-                var selectedTeam = privateData.teams[teamValue] || {};
-                team = {
-                    id: Number(teamValue),
-                    edPublic: selectedTeam.edPublic
-                };
+            if (teamValue && teamValue !== 'none') {
+                team = privateData.teams[teamValue] || {};
+                team.id = Number(teamValue);
             }
 
             return {
--- a/www/common/common-ui-elements.js
+++ b/www/common/common-ui-elements.js
@@ -2825,21 +2825,23 @@ define([
         var teamOptions = [{
             tag: 'a',
             attributes: { 'data-value': '-1' },
-            content: [Messages.settings_cat_drive]
+            content: [h('span.cp-creation-team-avatar', Icons.get('drive')), h('span.cp-creation-team-name', Messages.settings_cat_drive)]
         }];
         Object.keys(privateData.teams || {}).forEach(function (id) {
             var data = privateData.teams[id];
             if (!data) { return; }
+            var avatar = h('span.cp-creation-team-avatar.cp-avatar');
+            common.displayAvatar($(avatar), data.avatar, data.name);
             teamOptions.push({
                 tag: 'a',
                 attributes: { 'data-value': id },
-                content: [data.name]
+                content: [avatar, h('span.cp-creation-team-name', data.name)]
             });
         });
         teamOptions.push({
             tag: 'a',
             attributes: { 'data-value': 'none' },
-            content: [Messages.autostore_hide]
+            content: [h('span.cp-creation-team-avatar', Icons.get('close')), h('span.cp-creation-team-name', Messages.autostore_hide)]
         });
         var $teamSelect = UIElements.createDropdown({
             text: getTeamLabel(teamValue),
@@ -2858,12 +2860,14 @@ define([
         setTeamTitle(getTeamLabel(teamValue));
         $teamSelect.onChange.reg(function (text, value) {
             teamValue = value == null ? 'none' : String(value);
-            setTeamTitle(text || getTeamLabel(teamValue));
+            var label = getTeamLabel(teamValue);
+            $teamBtn.find('.cp-dropdown-button-title').text(label);
+            setTeamTitle(label);
         });
         team = h('div.cp-creation-teams', [
             h('span.cp-creation-store-label', Messages.team_pcsSelectLabel),
             $teamSelect[0],
-            createHelper('#', Messages.team_pcsSelectHelp)
+            // createHelper('#', Messages.team_pcsSelectHelp) - commented until the documentation is updated
         ]);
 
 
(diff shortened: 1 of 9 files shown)


www/common/sframe-common-outer.js
```diff
@@ -2231,6 +2231,14 @@ define([
                     });
                 });
             });
+            sframeChan.on('Q_GET_EDIT_URL', function (data, cb) {
```
yflory: You can't send the "edit" or "view" URL to the sandboxed iframe. If you want to display keys in the UI you must use the "secure iframe"

www/common/common-ui-elements.js
```diff
@@ -2811,55 +2811,79 @@ define([
...
         // * -1 ==> store in the user drive, and the user will be the owner
-        // * undefined ==> ask
-        if (teamExists) {
-            var teams = Object.keys(privateData.teams).map(function (id) {
-                var data = privateData.teams[id];
-                var avatar = h('span.cp-creation-team-avatar.cp-avatar');
-                // We assume that teams always have a non-empty name, so we don't need a UID
+        // * undefined ==> ask (CryptDrive / team / don't store); default CryptDrive
+        var team;
+        var teamValue = privateData.storeInTeam != null ? String(privateData.storeInTeam) : '-1';
```
yflory: `npm run lint`

yflory (changes requested):

www/common/common-ui-elements.js
```diff
@@ -2811,55 +2811,79 @@ define([
...
         // * -1 ==> store in the user drive, and the user will be the owner
-        // * undefined ==> ask
-        if (teamExists) {
-            var teams = Object.keys(privateData.teams).map(function (id) {
-                var data = privateData.teams[id];
-                var avatar = h('span.cp-creation-team-avatar.cp-avatar');
-                // We assume that teams always have a non-empty name, so we don't need a UID
+        // * undefined ==> ask (CryptDrive / team / don't store); default CryptDrive
+        var team;
+        var teamValue = privateData.storeInTeam !== null ? String(privateData.storeInTeam) : '-1';
```
yflory: `privateData.storeInTeam` seems to be `undefined` when it isn't a team ID, so checking against `null` won't work. As a result, the default action when it isn't a team ID is "undefined" and the document is not stored.

The clean way is probably to check if "storeInTeam" is undefined (we can't just check if it's falsy because  `0` is a valid ID).

www/common/common-ui-elements.js
```diff
@@ -2811,55 +2811,79 @@ define([
...
-                $team.find('[data-id="'+privateData.storeInTeam+'"]').addClass('cp-selected');
-                teamValue = privateData.storeInTeam;
             }
-        }
+            return avatar;
+        };
+        var teamOptions = [{
+            tag: 'a',
+            attributes: { 'data-value': '-1' },
+            content: [getTeamAvatar('-1'), h('span.cp-creation-team-name', Messages.settings_cat_drive)]
```
yflory: You can re-use `getTeamsLabel('-1')` here instead of using the translation key. It can avoid inconsistencies in the future if we change one value and not the other.

www/common/common-ui-elements.js
```diff
@@ -2811,55 +2811,79 @@ define([
...
+            teamOptions.push({
+                tag: 'a',
+                attributes: { 'data-value': id },
+                content: [getTeamAvatar(id), h('span.cp-creation-team-name', data.name)]
+            });
+        });
+        teamOptions.push({
+            tag: 'a',
+            attributes: { 'data-value': 'none' },
+            content: [getTeamAvatar('none'), h('span.cp-creation-team-name', Messages.autostore_hide)]
```
yflory: Same as the drive, you can re-use `getTeamsLabel()` here instead of the translation key.

www/common/common-ui-elements.js
```diff
@@ -2811,55 +2811,79 @@ define([
...
+        });
+        var $teamBtn = $teamSelect.find('button').addClass('btn');
+        var setTeamButton = function (val) {
+            var label = getTeamLabel(val);
+            $teamBtn[0]?._tippy?.destroy();
+            $teamBtn.attr('title', label);
+            $teamBtn.find('.cp-dropdown-button-title').empty().append([
+                getTeamAvatar(val),
+                h('span.cp-creation-team-name', label)
+            ]).removeAttr('title');
```
yflory: Why do you need `.removeAttr('title');` here?

www/common/common-ui-elements.js
```diff
@@ -2811,55 +2811,79 @@ define([
...
+        };
+        setTeamButton(teamValue);
+        $teamSelect.onChange.reg(function (text, value) {
+            teamValue = value === null ? 'none' : String(value);
```
yflory: I don't think the value can be `null` here. If there is an unexpected issue and something wrong with the DOM, it could be `undefined`.

www/common/sframe-common-outer.js
```diff
@@ -2231,6 +2231,15 @@ define([
                     });
                 });
             });
+            sframeChan.on('EV_FORM_GUEST_SHARE_OPEN', function (data) {
+                initSecureModal('formGuestShare', data || {}, function (action) {
+                    if (action && action.action === 'openView') {
+                        var url = Utils.Hash.hashToHref(hashes.viewHash, 'form');
+                        var a = window.open(url);
+                        if (!a) { sframeChan.event('EV_POPUP_BLOCKED'); }
+                    }
```
yflory: This code is duplicated from the "EV_OPEN_VIEW_URL" command.
You can move `sframeChan.on('EV_OPEN_VIEW_URL', function () {` from this file into the `addCommonRpc` section in order to be able to use it directly from the secure iframe.

www/form/inner.js
```diff
@@ -4908,14 +4906,36 @@ define([
                 Messages.form_geturl
             ]);
             var preview = h('div.cp-forms-results-participant', [previewBtn, participantBtn]);
+            var shouldWarnGuestAuthor = function () {
+                if (framework._.sfCommon.isLoggedIn()) { return false; }
+                if (!APP.isEditor) { return false; }
+                return true;
+            };
+            var warnGuestAuthorThen = function (next) {
+                if (!shouldWarnGuestAuthor()) { return void next(); }
```
yflory: This line seems to be useless since we already check `shouldWarnGuestAuthor` before calling `warnGuestAuthorThen`. And this is a good thing because it would create a TypeError (`next` is a String).

yflory (changes requested):
The move to the Secure Iframe has created a few small issues which *should be* easy to fix.
I've also found another issue while testing: the "Don't store" action from the pad creation screen doesn't work and pads are stored in my drive.

yflory (approved):

PR #2350, opened 2026-08-03
Enable form printing

### LINKED ISSUE #2024: Print/screenshot form as participant broken
### Description

It makes sense for participants to want to have their own copy of a form they've filled in. The easiest way to do this is to print the form (to PDF or paper). An alternative is to take a full-page screenshot.

However, the result is cut off after just one page, even if you explicitly choose to print the inner frame.

Tested on latest Chromium and Firefox based desktop browsers.

### Steps to reproduce

1. Open a form as a participant
2. Choose the browser's option to either print, or create a full-page screenshot

### Relevant logs

```shell

```

### Expected behavior

A screenshot or print of the entire form should be possible.

Microsoft Windows

Mozilla Firefox, Google Chrome, Microsoft Edge

### Extensions

N/A

### Version

2025.6.0

zuzanna-maria:
- fixes #2024 (can print entire form)


www/form/inner.js
```diff
@@ -127,6 +127,22 @@ define([
 
         return cancelBlock;
     };
+
+    var nativePrint = window.print;
+
+    var printFullContent = function () {
+        setTimeout(function () { nativePrint(); }, 50);
```
yflory: Why do you need a 50ms timeout?
This seems either unnecessary (and 0ms would be enough) or subject to race condition (and we'd need a better solution than a timeout).

yflory (changes requested):

yflory (approved):

PR #2356, opened 2026-08-05
Enable printing code documents in embed mode

### LINKED ISSUE #403: Code: printing of embed/present/ not possible
The print stylesheet hides the element `#cp-app-code-preview-content` but without using the print button in the toolbar (which is hidden by using embed/present) the element `#cp-app-code-print` (which is filled by the print button) keeps being empty.

This results in an empty page and an empty printout for the user.

zuzanna-maria:
- fixes #403 (can print Code documents shared in embed mode using ctrl+p)

### DIFF
--- a/www/code/inner.js
+++ b/www/code/inner.js
@@ -522,20 +522,10 @@ define([
             mkHelpMenu(framework);
         }
 
-        var nativePrint = window.print;
-
-        var printFullContent = function () {
-            var $print = $('#cp-app-code-print');
-            var $content = $('#cp-app-code-preview-content');
-            $print.html($content.html());
-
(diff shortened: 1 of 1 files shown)


yflory (changes requested):
The current code should work but you don't have to re-implement the code of the print button. In Embed mode, the "File" menu exists but it is hidden. You should be able to call directly (or store the button in a variable) when using "Ctrl+P":
```js
$('.cp-toolbar-icon-print').click() ;
```

You could also probably remove the `privateData.isPresent` check to make sure the keyboard shortcut also works in non-embed mode.

yflory (approved):

Issue #2368, opened 2026-08-28
Read-only filesystem for docker image


wginolas:
- Add [`read_only: true`](https://docs.docker.com/reference/compose-file/services/#read_only) to the `docker-compose.yml`
- Set `readOnlyRootFilesystem: true` here: https://github.com/cryptpad/helm/blob/main/charts/cryptpad/values.yaml#L177
- Make sure CryptPad still works with this setting.
  - If it does not work, we may need to add `tmpfs` or other volumes.

wginolas:
There is a duplicate of this ticket: https://github.com/cryptpad/cryptpad/issues/1313

gsautner:
Created MR on opendesk side: https://gitlab.opencode.de/bmi/opendesk/deployment/opendesk/-/merge_requests/1508

On my tests, just adding this extra volumes map was sufficient to work (with the [version](https://github.com/cryptpad/helm/releases/tag/cryptpad-helm-0.1.0) of Helm made by the [contributor](https://github.com/cryptpad/helm/pull/42)): 
```
extraVolumes:
  - name: "customize"
    emptyDir: {}
  - name: "npm-cache"
    emptyDir: {}

extraVolumeMounts:
  - name: "customize"
    mountPath: "/cryptpad/customize"
  - name: "npm-cache"
    mountPath: "/cryptpad/.npm"
```

wginolas:
Thanks!

Issue #1313, opened 2023-11-08
Support run Docker image with a Read-Only Filesystem


gsautner:
### Description

The idea is to run CryptPad container image on very secure infrastructure that enforces a read-only file system for container image files. 

### Steps to reproduce

Was used [CryptPad/Helm](https://github.com/cryptpad/helm) chart to deploy an instance with these values: 
```
enableEmbedding: true 
workloadStateful: false
persistence:
  enabled: false
securityContext: 
  capabilities:
    drop:
     - ALL
  readOnlyRootFilesystem: true
```

### Relevant logs

```shell
npm ERR! Log files were not written due to an error writing to the directory: /cryptpad/.npm/_logs
npm ERR! You can rerun the command with `--loglevel=verbose` to see the logs in your terminal
Creating ./www/sheet/index.html
/cryptpad/node_modules/fs-extra/lib/mkdirs/mkdirs-sync.js:45
        throw err0
        ^

Error: EROFS: read-only file system, mkdir '/tmp/CRYPTPAD_TEMP_BUILD'
    at Object.mkdirSync (node:fs:1398:3)
    at mkdirsSync (/cryptpad/node_modules/fs-extra/lib/mkdirs/mkdirs-sync.js:31:9)
    at mkdirsSync (/cryptpad/node_modules/fs-extra/lib/mkdirs/mkdirs-sync.js:36:14)
    at Object.mkdirsSync (/cryptpad/node_modules/fs-extra/lib/mkdirs/mkdirs-sync.js:36:14)
    at write (/cryptpad/scripts/build.js:120:9)
    at /cryptpad/scripts/build.js:231:5
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/cryptpad/scripts/build.js:206:19)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10) {
  errno: -30,
  syscall: 'mkdir',
  code: 'EROFS',
  path: '/tmp/CRYPTPAD_TEMP_BUILD'
}

Node.js v18.18.1
```

### Expected behavior

Some alternative to setup custom working directory or another temporary folder that doesn't contain files on image.  

Or maybe a list of all directories that can be mounted to support this read only execution. 

### What opreating system are you using?

Other

### Extensions

N/A

### Version

5.5.0

PR #2383, opened 2026-09-16
Run CryptPad Docker in an read-only file system

wginolas:
https://github.com/cryptpad/cryptpad/issues/2368
https://github.com/cryptpad/cryptpad/issues/1313


wginolas:
1. `config` is controlled by git and should not be chowned
2. We are not happy how CryptPad is configured, and this PR makes it worse. (docker-compose env vars vs. config.js)

I'll close this for now and think of a better way to do this.

PR #2389, opened 2026-09-21
Set diagram UI theme to internal theme id, not localized theme label

xenofem:


davidbenque:
Hi and thank you for your contribution. Can you clarify what you are trying to achieve with this PR? is it to display the theme ID in the interface instead of a localized name?

xenofem:
I'm not trying to change anything about the interface; it's just that the theme selector is currently broken because it doesn't actually use the values `"sketch"`, `"simple"`, or `"classic"` internally. When I select "Sketch" in English, it tries to internally set the Diagrams UI theme to the value `"Sketch"` with a capital S, which isn't a valid option, so Diagrams just reverts to the Classic theme because it doesn't know what else to do. If I had the UI set to Chinese, it would try to set the Diagrams UI theme to the value `"草图"`, which would also be invalid. The localized strings should be used *only* in the UI, not as parameters for the internal configuration of the Diagrams tool.

xenofem:
This didn't come up as an issue when the theme selector was first implemented in #2192 because at that point in time, the UI message strings happened to *also* be `"sketch"`, `"simple"`, and `"classic"`, and there weren't any translations provided yet. When the messages were updated to use capitalized words and translations were added, the feature quietly broke.

